PR #77, opened 2014-12-19
Fix title and rel attributes in markdown links

steventux:
Removes markdown for `title` and `rel` link attributes from all published `TravelAdviceEdition`s.
https://www.agileplannerapp.com/boards/173808/cards/9021

### DIFF
--- a/db/migrate/20141217122045_remove_link_title_attributes.rb
+++ b/db/migrate/20141217122045_remove_link_title_attributes.rb
@@ -1,0 +1,60 @@
+class RemoveLinkTitleAttributes < Mongoid::Migration
+  LINK_REGEX = %r{
+                (\[.*?\])           # link text in literal square brackets
+                \(                # literal opening parenthesis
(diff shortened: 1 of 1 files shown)


db/migrate/20141217122045_remove_link_title_attributes.rb
```diff
@@ -0,0 +1,60 @@
...
+    str.gsub(/(“|”)+/, '"').gsub(LINK_REGEX) do |link|
+      link_text_md  = $1
+      link_url      = $2
+      link_title    = $3
+      link_rel      = $4
+
+      prepend_scheme = link_url =~ /\Awww/
+
+      if link_title or link_rel or prepend_scheme
+        link_url = "http://#{link_url}" if prepend_scheme
```
steventux: I used very recent data when working on this and in all cases the URLs were `http` (there were only a handful of incomplete URLs like this).

PR #79, opened 2015-01-27
Remove setting of per-country related artefacts.

danielroseman:
This has never been used, and doesn't currently work due to the volume of data in the artefacts collection. Removing it simplifies the migration to content-store.This has never been used and complicates the migration to content-store.


Gemfile.lock
```diff
@@ -102,7 +102,7 @@ GEM
       state_machine
     hashie (3.2.0)
     hike (1.2.3)
-    htmlentities (4.3.1)
+    htmlentities (4.3.3)
```
alext: Why this change?

alext:
Other than the `Gemfile.lock` change, this all looks good.  Will make the migration a lot easier.

alext:
It's also worth mentioning that this feature doesn't currently work anyway due to the volume of data in the artefacts collection.

danielroseman:
Reverted lock file, updated commit comment and force pushed.

PR #80, opened 2015-06-17
Patch Rails to fix security vulnerability

binaryberry:
Bump Rails gem to 3.2.22 to fix
https://groups.google.com/forum/#!msg/rubyonrails-security/bahr2JLnxvk/x4EocXnHPp8J

Bumped govuk_admin_template gem version to 2.3.1 to fix
jquery-rails vulnerability

Added initializer to fix:
https://groups.google.com/forum/#!topic/rubyonrails-security/7VlB_pck3hU
(which is not fixed by the upgrade to rails 3.2.22)


edds:
Can the monkey patch be updated to include the 'raise' from this: https://github.com/alphagov/static/pull/605/files so that we will know to remove it in the future.

binaryberry:
Sure @edds - done now!

PR #84, opened 2015-09-15
Send content_ids to panopticon.

alext:
This adds content_ids, and includes them in the data sent to Panopticon.

Sending these to publishing-api will be done in a follow-up PR.

### DIFF
--- a/spec/lib/registerable_travel_advice_edition_spec.rb
+++ b/spec/lib/registerable_travel_advice_edition_spec.rb
@@ -60,4 +60,21 @@ describe RegisterableTravelAdviceEdition do
       @registerable.prefixes.should == ["/foreign-travel-advice/#{@edition.country_slug}"]
     end
   end
+
+  describe "content_id" do
+    before :each do
+      @edition = FactoryGirl.build(:travel_advice_edition)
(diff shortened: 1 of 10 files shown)


spec/lib/registerable_travel_advice_edition_spec.rb
```diff
@@ -60,4 +60,21 @@
...
+
+  describe "content_id" do
+    before :each do
+      @edition = FactoryGirl.build(:travel_advice_edition)
+      @registerable = RegisterableTravelAdviceEdition.new(@edition)
+    end
+
+    it "should return the content_id of the corresponding country" do
+      @edition.country_slug = 'albania'
+      @registerable.content_id.should == Country.find_by_slug('albania').content_id
```
rboulton: This test feels circular - the right hand side is almost exactly the same code as the implementation of the LHS runs.

Might it be more useful to use the literal expected value on the RHS (with a `# from countries.yml fixture` comment, like in `edition_edit_spec.rb`)?

alext: Good call.  I've updated this as part of #85.

PR #85, opened 2015-09-17
Send placeholder items to publishing-api

alext:
To enable us to keep tagging in sync between old and new world.


bishboria:
:+1:

PR #91, opened 2015-12-21
Publish to publishing api v2

tuzz:
I have tested a publish into the frontend application locally on my machine. Prior to deploy, we should repeat this test on preview/staging.


jackscotti:
You'll need to update this to test against v2 content schemas:
https://github.com/alphagov/travel-advice-publisher/blob/master/spec/support/govuk_content_schemas.rb#L4

jackscotti:
:100: * better than I would have done! Added a couple of comments, looks good.

tuzz:
I think the content schema configuration is correct @jackscotti. It's using the publisher/ schemas inside travel_advice_publisher: https://github.com/alphagov/govuk-content-schemas/tree/master/formats/travel_advice/publisher

The presenter tests explicitly state which format to validate against:
- https://github.com/alphagov/travel-advice-publisher/pull/91/files#diff-8180fe08951fde62fc6dd9c41bee0d17R50
- https://github.com/alphagov/travel-advice-publisher/pull/91/files#diff-65e2fd0ef9c96bdd879dd5393aa3c0f6L41

tuzz:
I have addressed all comments. Please could you take another look. @danielroseman  @jackscotti

jackscotti:
Looks good to me. But I am not familiar with the placeholder change @danielroseman is talking about.

PR #98, opened 2016-01-13
Update poltergeist to 1.8.1...

steventux:
Travel advice publisher [tests have been failing for a while but passing CI](https://ci.dev.publishing.service.gov.uk/job/govuk_travel_advice_publisher_branches/89/console) because the suite has ended with an exit code of `0`. This is [traceable to a bug with an `at_exit` call in Poltergeist](https://github.com/teampoltergeist/poltergeist/issues/497).
[Updating poltergeist fixes this](https://github.com/teampoltergeist/poltergeist/commit/f1b894ee71578806cd52415a89fd52498051deeb) and [breaks the build](https://ci.dev.publishing.service.gov.uk/job/govuk_travel_advice_publisher_branches/108/console ) woo!
I'll fix that next...


benlovell:
Merging despite the build failing, as that's the point. :wink:

PR #101, opened 2016-01-19
Adds static 'breadcrumbs' data for countries.

steventux:
https://trello.com/c/vm54jvVo/477-send-hard-coded-breadcrumbs-to-publishing-api

also relates to https://github.com/alphagov/govuk-content-schemas/pull/205 and the accompanying story https://trello.com/c/tomHUlp7/475-define-data-format-for-breadcrumbs

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -51,6 +51,7 @@ private
       "email_signup_link" => TravelAdvicePublisher::EMAIL_SIGNUP_URL,
       "parts" => parts,
       "alert_status" => edition.alert_status,
+      "primary_parent" => breadcrumbs,
     }
 
     details.merge!("image" => image) if image
@@ -98,5 +99,9 @@ private
 
   def document
     @document ||= AssetPresenter.present(edition.document)
+  end
+
+  def breadcrumbs
+    @breadcrumbs ||= BreadcrumbsPresenter.present
   end
 end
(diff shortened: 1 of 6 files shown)


spec/presenters/breadcrumbs_presenter_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "spec_helper"
+
+RSpec.describe BreadcrumbsPresenter do
```
benlovell: You could collapse this and the next line into `RSpec.describe BreadcrumbsPresenter, ".present" do`

danielroseman:
:+1:

app/presenters/edition_presenter.rb
```diff
@@ -51,6 +51,7 @@ def details
       "email_signup_link" => TravelAdvicePublisher::EMAIL_SIGNUP_URL,
       "parts" => parts,
       "alert_status" => edition.alert_status,
+      "primary_parent" => breadcrumbs,
```
rboulton: In the current schemas for links, we're using the `parent` field for the primary parent, and I remember having discussions suggesting that we add `parents` or `all_parents` for holding the list of all parents.  See https://github.com/alphagov/govuk-content-schemas/blob/master/formats/base_links.json#L21

This doesn't conflict with this PR, since this PR is putting something into the details hash, but would it be worth trying to make the terminology consistent at this point to avoid future confusion.  Either by changing this PR to use `parent`, or by changing the links "base_links" schema to use a link type `primary_parent` (and presumably also adding an `all_parents` link to the schema, too).

steventux:
Closing as publishing platform will revisit this shortly.

PR #106, opened 2016-01-27
Add static 'breadcrumbs' data to travel advice...

steventux:
https://trello.com/c/vm54jvVo/477-send-hard-coded-breadcrumbs-to-publishing-api
Static breadcrumbs for travel advice country and index representations sent to the publishing api in the links hash.

### DIFF
--- a/app/presenters/breadcrumbs_presenter.rb
+++ b/app/presenters/breadcrumbs_presenter.rb
@@ -1,0 +1,39 @@
+class BreadcrumbsPresenter
+  def self.present
+    [
+      {
+        "content_id" => "08d48cdd-6b50-43ff-a53b-beab47f4aab0",
+        "base_path" => "/foreign-travel-advice",
+        "title" => "Foreign travel advice",
+        "links" => { "parent" => ["b9849cd6-61a7-42dc-8124-362d2c7d48b0"] },
+      },
+      {
+        "content_id" => "b9849cd6-61a7-42dc-8124-362d2c7d48b0",
+        "base_path" => "/browse/abroad/travel-abroad",
+        "title" => "Travel abroad",
+        "links" => { "parent" => ["86eb717a-fb40-42e7-83fa-d031a03880fb"] },
+      },
+      {
+        "content_id" => "86eb717a-fb40-42e7-83fa-d031a03880fb",
+        "base_path" => "/browse/abroad",
+        "title" => "Passports, travel and living abroad",
+      }
+    ]
+  end
+
+  def self.present_for_index
+    [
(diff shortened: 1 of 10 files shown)


app/presenters/breadcrumbs_presenter.rb
```diff
@@ -0,0 +1,39 @@
...
+        base_path: "/browse/abroad/travel-abroad",
+        title: "Travel abroad",
+        links: { parent: ["86eb717a-fb40-42e7-83fa-d031a03880fb"] },
+      },
+      {
+        content_id: "86eb717a-fb40-42e7-83fa-d031a03880fb",
+        base_path: "/browse/abroad",
+        title: "Passports, travel and living abroad",
+      }
+    ].as_json
```
elliotcm: The API client will do this, won't it?

app/presenters/index_links_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+        :links => {
+          "parent" => breadcrumbs
+        }
+      }
+    end
+
+  private
+
+    def breadcrumbs
+      @breadcrumbs ||= BreadcrumbsPresenter.present_for_index
```
elliotcm: Given the sometimes unpredictable nature of class-level memoisation in Ruby do we need this?  The `BreadcrumbsPresenter` is only building a static data structure, and doing it infrequently.

elliotcm: That would also allow us to inline the call to the presenter and remove the `class << self` declaration which is frowned upon in the [styleguide](https://github.com/alphagov/styleguides/blob/master/ruby.md#classes).

app/presenters/links_presenter.rb
```diff
@@ -0,0 +1,29 @@
+class LinksPresenter
+
+  attr_reader :edition
```
elliotcm: This reader can be private.

spec/presenters/index_presenter_spec.rb
```diff
@@ -82,8 +83,8 @@
                 "change_description" => "Stuff changed",
                 "synonyms" => ["foo", "bar"],
               },
-            ],
-          },
+            ]
+          }
```
elliotcm: Trailing commas are recommended to keep diffs cleaner when adding/removing entries: https://github.com/alphagov/styleguides/blob/master/ruby.md#collections

elliotcm:
:+1: but for a few style/class questions (not that kind of style or class)

steventux:
@elliotcm thanks for review, updated.

PR #107, opened 2016-01-27
Email alert integration

tuzz:
Can be merged safely. The email alert api integration is guarded by a feature flag.

https://trello.com/c/JHXCib9K/512-trigger-emails-from-travel-advice-publisher

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -80,6 +80,7 @@ class Admin::EditionsController < ApplicationController
       PublishingApiNotifier.put_links(@edition)
       PublishingApiNotifier.publish(@edition)
       PublishingApiNotifier.publish_index
+      EmailAlertApiNotifier.send_alert(@edition)
 
       # catch any upload errors
       if @edition.errors.any?
--- a/app/presenters/email_alert_presenter.rb
+++ b/app/presenters/email_alert_presenter.rb
@@ -1,0 +1,88 @@
+class EmailAlertPresenter
+  def self.present(edition)
+    new(edition).present
+  end
+
+  def initialize(edition)
+    self.edition = edition
+  end
+
+  def present
+    {
+      "subject" => subject,
+      "tags" => tags,
+      "links" => links,
+      "format" => format,
+      "body" => body,
+    }
+  end
+
+private
+  attr_accessor :edition
+
+  def subject
+    edition.title
+  end
+
+  def tags
+    {}
+  end
+
+  def links
+    { countries: [content_id] }
+  end
+
+  def format
+    "placeholder_travel_advice"
+  end
+
+  def body
+    <<-HTML
+      <div class="rss_item" data-message-id="#{message_identifier}" style="margin-bottom: 2em;">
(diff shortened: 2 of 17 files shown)


app/presenters/email_alert_presenter.rb
```diff
@@ -0,0 +1,75 @@
...
+    edition.title
+  end
+
+  def tags
+    { countries: [edition.country_slug] }
+  end
+
+  def links
+    :TODO
+  end
```
tuzz: I need to discuss this with Finding Things.

app/controllers/admin/editions_controller.rb
```diff
@@ -78,6 +78,7 @@ def save_and_publish
       PublishingApiNotifier.put_content(@edition)
       PublishingApiNotifier.publish(@edition)
       PublishingApiNotifier.publish_index
+      EmailAlertApiNotifier.send_alert(@edition)
```
danielroseman: The list of API calls that happen on save is getting quite long. I'm hesitant to suggest adding sidekiq but I do worry about what would happen if one of this chain times out, for whatever reason.

tuzz: Yeah – it's not great. At least we'll only send out an email if all over the other actions completed.

app/presenters/email_alert_presenter.rb
rboulton: This story (at the top of our backlog) is relevant to this discussion, I believe: https://trello.com/c/gdrGvo9n/501-update-email-alert-service-to-support-email-notifications-for-items-tagged-to-topics-policies-in-their-links-hash

rboulton: Ah - actually, this isn't relevant, because I was expecting this email sending to be using email-alert-service, but it's going directly to email-alert-api instead.  I'm not sure that's path is desirable - but I suppose that's part of the discussion you want to have with us, so I'll leave it for offline discussion.

elliotcm: After in-person discussion, this is how we plan on dealing with the question of which links to use here and how to prevent other document types from triggering travel alerts: https://trello.com/c/MEBIJSDd/526-make-email-alert-api-check-formats

@rboulton could you check and confirm I've written that up correctly?

rboulton: That writeup looks good to me.

elliotcm:
:+1: though we may have to restructure the request body based on some Rails annoyances in the email alert API.

app/presenters/email_alert_presenter.rb
```diff
@@ -0,0 +1,88 @@
...
+  def initialize(edition)
+    self.edition = edition
+  end
+
+  def present
+    {
+      "subject" => subject,
+      "tags" => tags,
+      "links" => links,
+      "format" => format,
```
danielroseman: After discussion with Richard, we've agreed to call this field `document_type`.

danielroseman:
:+1: apart from the field name change from format to document_type.

tuzz:
Rename is complete. Merge when happy.

PR #112, opened 2016-02-09
Upgrade to Rails 4.2 and Mongoid 5

danielroseman:
This relies on the upgrade to govuk_content_models which is currently on a branch; when all apps have been upgraded we will need to do a release and update the gemfiles.

This also relies on changes to govuk-puppet and deployment to add the new SECRET_KEY_BASE and MONGODB_URI environment variables.


tuzz:
:+1: All looks sensible.

benlovell:
:sparkler: :+1: Worth adding the linter separately and rebasing.

PR #115, opened 2016-02-16
Explicitly include test-unit Gem everywhere

bradwright:
ActiveSupport appears to depends on test-unit, which was removed in 2.2.0:

https://www.ruby-lang.org/en/news/2014/12/25/ruby-2-2-0-released/

Adding this back in at the top level allows us to get a Rails console again.

This should only be an issue until [we go to Rails 4](https://github.com/alphagov/travel-advice-publisher/pull/112) - it appears to be [related to `3.2.x`](https://github.com/rspec/rspec-rails/issues/1273#issuecomment-70681180).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,8 +52,10 @@ group :development, :test do
   gem 'timecop', '0.5.9.2'
   gem 'jasmine', '2.1.0'
   gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
-  gem 'test-unit'
   gem 'govuk-lint', '~> 0.5.3'
   gem 'pry-rails'
   gem "pry-byebug"
 end
+
+# FIXME: move back into the `test` group once we're on Rails 4
+gem 'test-unit'


Gemfile
```diff
@@ -53,6 +53,7 @@ group :development, :test do
   gem 'timecop', '0.5.9.2'
   gem 'jasmine', '2.1.0'
   gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
-  gem 'test-unit'
   gem 'pry-rails'
 end
+
+gem 'test-unit'
```
jamiecobbett: I think this should be commented to say that it should be moved into the test group once this app is migrated to Rails 4.

bradwright: Fixed and force pushed

bradwright:
Going to add those links to the commit - don't merge yet

bradwright:
Ready to merge,.

PR #121, opened 2016-03-02
Email alert signup

tuzz:
https://trello.com/c/oqGCU2XP/525-add-email-signup-links-to-travel-advice-on-frontend

Depends on https://github.com/alphagov/govuk-content-schemas/pull/257

screen shot 2016-03-02 at 16 39 57

screen shot 2016-03-02 at 16 39 46

### DIFF
--- a/app/presenters/email_alert_signup/edition_presenter.rb
+++ b/app/presenters/email_alert_signup/edition_presenter.rb
@@ -1,0 +1,82 @@
+module EmailAlertSignup
+  class EditionPresenter
+    def initialize(edition)
+      self.edition = edition
+    end
+
+    def content_payload
+      {
+        content_id: content_id,
+        base_path: base_path,
+        format: "email_alert_signup",
+        title: edition.title,
+        description: "#{edition.title} Email Alert Signup",
+        public_updated_at: public_updated_at,
+        locale: "en",
+        publishing_app: "travel-advice-publisher",
+        rendering_app: "email-alert-frontend",
+        routes: routes,
+        details: details,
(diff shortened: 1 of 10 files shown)


elliotcm:
@tuzz and I discussed the "republish" type and the fact that this uses a rake task rather than republishing on the fly like policy publisher does.  Because the country names are rarely (if ever) edited and same for new countries being created, it seems find to have this as rake tasks.

Because now two things need to be done if you change a country name or add a new one, he's adding a section to the README.

app/presenters/email_alert_signup/edition_presenter.rb
```diff
@@ -0,0 +1,82 @@
...
+      {
+        signup_tags: tags,
+        summary: summary,
+        breadcrumbs: breadcrumbs,
+        govdelivery_title: edition.title,
+      }
+    end
+
+    def tags
+      { countries: [country.content_id] }
```
bradwright: Does this also need `document_format`?

elliotcm: Not in that set of tags but yeah it'll have to go somewhere.

elliotcm:
:+1: when green

tuzz:
I've added some tasks to the ticket. One of which is to update email-alert-frontend so that it passes the `document_type` through if present.

Nothing will link to these pages until we update the subscription links on the index/edition content items for the countries. So going to merge this now (@elliotcm said he was happy for this to be self-merged).

Will move onto email-alert-frontend tomorrow.

PR #126, opened 2016-03-29
Cache travel advice content for a maximum of 10 seconds

tuzz:
These tests are failing because of the schemas, which are updated here:
https://github.com/alphagov/govuk-content-schemas/pull/273


danielroseman:
:+1:

tuzz:
@danielroseman tests are now passing as schemas has been merged

PR #127, opened 2016-03-30
Enqueue publishing API calls

elliotcm:
https://trello.com/c/ABfCJwT7/648-travel-advice-publisher-send-content-on-a-queue-rather-than-synchronously

This code by @steventux makes our publishing API calls async to provide additional resilience to publishing actions.


danielroseman:
Looks good, although the thing that might worry me about this is the amount of data we're putting into a single payload. Could that be a concern with sidekiq/redis?

elliotcm:
Redis is bound only by memory, so it should be OK, certainly it didn't cause any issues in the small-scale testing I did locally with the Afghanistan location.

PR #128, opened 2016-03-30
Trigger pub API index update in separate worker.

elliotcm:
This should prevent the (less important) update of
the travel advice index page from affecting updates
to the individual travel advice pages if the index
page update fails or conflicts.

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -97,9 +97,13 @@ class Admin::EditionsController < ApplicationController
       notifier.put_content(@edition)
       notifier.put_links(@edition)
       notifier.publish(@edition)
-      notifier.publish_index
       notifier.enqueue
+
       EmailAlertApiNotifier.send_alert(@edition)
+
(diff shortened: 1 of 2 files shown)


app/controllers/admin/editions_controller.rb
```diff
@@ -97,10 +97,14 @@ def save_and_publish
       notifier.put_content(@edition)
       notifier.put_links(@edition)
       notifier.publish(@edition)
-      notifier.publish_index
       notifier.enqueue
+
       EmailAlertApiNotifier.send_alert(@edition)
```
danielroseman: just noticed that this is not in the queue at all, which means it can happen before the publish. Could that be a problem?

danielroseman:
Probably worth an test to ensure that save+publish triggers two separate notifier tasks.

app/controllers/admin/editions_controller.rb
elliotcm: Our conclusion after an in-person chat was "probably fine".

elliotcm:
@danielroseman Tests updated.

PR #131, opened 2016-03-31
Add rake task to trigger email alerts for an edition.

elliotcm:
https://trello.com/c/Xr0LF1IP/653-rake-resend-alert-task-for-ta-advice

Opsmanual update: https://github.gds/gds/opsmanual/pull/659

### DIFF
--- a/lib/tasks/email_alerts.rake
+++ b/lib/tasks/email_alerts.rake
@@ -1,0 +1,10 @@
+namespace :email_alerts do
+  desc "Triggers an email notification for the given edition ID"
+  task :trigger, [:edition_id] => :environment do |task, args|
+    abort "Please provide an edition ID, e.g. rake #{task.name}[fedc13e231ccd7d63e1abf65]" unless args[:edition_id]
+
+    edition = TravelAdviceEdition.find(id: args[:edition_id])
+    puts "Sending an email alert for #{edition.title}"
+    EmailAlertApiNotifier.send_alert(edition)
+  end
+end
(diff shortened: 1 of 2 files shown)


danielroseman:
Is there any way we can warn that this will potentially email hundreds of people? Or should that go in the opsmanual docs?

elliotcm:
@danielroseman If they're running it manually we could add a y/N confirmation, if they're triggering via jenkins then we can't.

elliotcm:
@danielroseman Also thousands is more likely.

lib/tasks/email_alerts.rake
```diff
@@ -0,0 +1,10 @@
+namespace :email_alerts do
+  desc "Triggers an email notification for the given edition ID"
+  task :trigger, [:edition_id] => :environment do |task, args|
+    abort "Please provide an edition ID, e.g. rake #{task.name}[fedc13e231ccd7d63e1abf65]" unless args[:edition_id]
+
+    edition = TravelAdviceEdition.find(id: args[:edition_id])
+    puts "Sending an email alert for #{edition.title}"
+    EmailAlertApiNotifier.send_alert(edition)
+  end
+end
```
tuzz: It's a bit pedantic, but could we provide a friendly message if they enter an id that doesn't exist. It's likely people unfamiliar with Ruby will be running this task.

tuzz:
:+1:

lib/tasks/email_alerts.rake
elliotcm: I considered that, but I checked the mongoid error and it's actually more detailed than anything I'd write.

tuzz: Ahh ok.

PR #132, opened 2016-04-04
Update README to reflect recent sidekiq changes

steventux:
The README contains references to the content API and doesn't accurately reflect the current production state of this application.
I've also added mention of sidekiq-monitoring [which should reach production soon](https://github.com/alphagov/sidekiq-monitoring/pull/17).


danielroseman:
Thanks for this. I think it would be worth adding something about the email alerts and how they work, as well.

steventux:
@danielroseman good point, I've added the relevant section.

PR #134, opened 2016-04-06
Enqueue email alerts with Sidekiq after a successful publish

tuzz:
https://trello.com/c/gwGF7DA8/666-enqueue-travel-advice-email-alert-with-sidekiq

The commits are fairly descriptive if you need more context of specific changes.

### DIFF
--- a/app/notifiers/publishing_api_notifier.rb
+++ b/app/notifiers/publishing_api_notifier.rb
@@ -40,6 +40,7 @@ class PublishingApiNotifier
   end
 
   def enqueue
+    validate_tasks_order
     worker.perform_async(tasks) if tasks.any?
   end
 
@@ -53,6 +54,21 @@ private
 
   def send_alert?(edition)
     edition.state == "published" && !edition.minor_update
+  end
+
+  def validate_tasks_order
+    endpoints = tasks.map(&:first)
+    send_alert_count = endpoints.count { |e| e == :send_alert }
+
+    if send_alert_count > 1
+      message = "send_alert must not be called more than once"
+      raise EnqueueError, message
+    end
+
+    if send_alert_count == 1 && endpoints[-2..-1] != [:publish, :send_alert]
+      message = "send_alert must be last and immediately follow a publish"
+      raise EnqueueError, message
+    end
   end
 
   class EnqueueError < StandardError; end
--- a/app/notifiers/publishing_api_notifier.rb
+++ b/app/notifiers/publishing_api_notifier.rb
@@ -31,8 +31,16 @@ class PublishingApiNotifier
     tasks << [:publish, presenter.content_id, presenter.update_type]
   end
 
+  def send_alert(edition)
+    return unless send_alert?(edition)
+
+    presenter = EmailAlertPresenter.new(edition)
+    payload = presenter.present
+    tasks << [:send_alert, presenter.content_id, payload]
+  end
+
   def enqueue
-    worker.perform_async(tasks)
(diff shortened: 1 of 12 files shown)


boffbowsh:
Overall, 👍  but can you squash some commits please? `check_email_alert_happens_last` is introduced and replaced in this pull request

app/workers/worker_error.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+private
+
+  def filter_backtrace(backtrace)
+    backtrace.select { |l| l.include?("travel-advice-publisher") }
+  end
+
+  def queue_size
+    queue = Sidekiq::Queue.all.first
```
danielroseman: While it's useful to have this (and the retry_set size) in the error log, is there a possibility that since it's calling out to an API it could fail? What would happen to the error message if it did?

tuzz: Good spot. There is that chance and that would cause the error message generation to exception and we'd get something cryptic in Errbit.

I will wrap these methods in a `rescue` to mitigate this risk.

tuzz:
@boffbowsh I have cleaned up the commit messages. I made it so that batch check is atomically introduced in the latter commit, which had a descriptive commit message.

@danielroseman I have added a blanket raise to the methods that add Sidekiq information to the error message. This is currently a blanket raise, which might be overly zealous, but seemed safest.

Please could you both take another look. Thanks.

boffbowsh:
👍 from me

tuzz:
I think @danielroseman will be busy for the rest of the day. Please could @elliotcm or @dougdroper take a look?

app/workers/worker_error.rb
dougdroper: Is this somehow different to `queue = Sidekiq::Queue.first`?

app/notifiers/publishing_api_notifier.rb
```diff
@@ -42,4 +51,25 @@ def enqueue
   def worker
     PublishingApiWorker
   end
+
+  def send_alert?(edition)
+    edition.state == "published" && !edition.minor_update
+  end
+
+  def check_tasks_are_ordered_sensibly
```
dougdroper: this method feels more like `validate_tasks_order` but not a big deal

app/notifiers/publishing_api_notifier.rb
```diff
@@ -31,8 +31,17 @@ def publish_index
     tasks << [:publish, presenter.content_id, presenter.update_type]
   end
 
+  def send_alert(edition)
+    if send_alert?(edition)
```
dougdroper: I think a prefer guards rather. `return if send_alert?(edition)` but also not a big deal.

app/workers/worker_error.rb
tuzz: `Sidekiq::Queue` doesn't respond to `.first`. It only provides an `.all` for whatever reason.

dougdroper: Ah well :( :+1:

app/notifiers/publishing_api_notifier.rb
tuzz: 👍  updated

app/notifiers/publishing_api_notifier.rb
tuzz: 👍  updated

tuzz:
@dougdroper said he was happy for this to be merged after rebase, which is now done.

PR #137, opened 2016-04-14
Add doc for the publish and notification workflow

pmanrubia:
Following the work on [Trello Card](https://trello.com/c/hW86qiR0/17-review-tap-alert-queueing) (paired with @gpeng), this PR include an initial set of notes of the `publish` and `notification` workflow.

### DIFF
--- a/docs/publish_and_notification_workflow.md
+++ b/docs/publish_and_notification_workflow.md
@@ -4,7 +4,7 @@ The act of ‘publishing’ an edition initiates a job with an ordered sequence 
 
 The [tasks order is validated](https://github.com/alphagov/travel-advice-publisher/blob/master/app/notifiers/publishing_api_notifier.rb#L43-43) to ensure that the email notification is always the final one.
 
-The last task (delivery of email via the email-alert-api) queues up a new Sidekiq job only if the previous tasks completed successfully (publising has succeded).
+The last task (delivery of email via the email-alert-api) queues up a new Sidekiq job only if the previous tasks completed successfully (publishing has succeeded).
 
 ## The Workflow
 The publish action [builds an array of four tasks](https://github.com/alphagov/travel-advice-publisher/blob/master/app/controllers/admin/editions_controller.rb#L97-L101).
--- a/docs/publish_and_notification_workflow.md
+++ b/docs/publish_and_notification_workflow.md
@@ -1,0 +1,22 @@
+# Publish and Notification Workflow
+
+The act of ‘publishing’ an edition initiates a job with an ordered sequence of tasks that send the content to the [publishing api](https://github.com/alphagov/publishing-api) and trigger an email alert via the [email alerts api](https://github.com/alphagov/email-alert-api).
+
+The [tasks order is validated](https://github.com/alphagov/travel-advice-publisher/blob/master/app/notifiers/publishing_api_notifier.rb#L43-43) to ensure that the email notification is always the final one.
+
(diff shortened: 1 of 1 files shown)


docs/publish_and_notification_workflow.md
```diff
@@ -0,0 +1,20 @@
+# Publish and Notification Workflow
+
+The act of ‘publishing’ an edition initiates an ordered sequence of jobs that send the content to the [publishing api](https://github.com/alphagov/publishing-api) and trigger an email alert via the [email alerts api](https://github.com/alphagov/email-alert-api).
```
boffbowsh: Are the jobs chained? Does the execution of one depend on the success of the other? Or are they just enqueued all together but in a prescribed order?

gpeng: They're passed into the worker in an array that has been ordered (and validated for content and order) prior to passing it in.

gpeng: In here https://github.com/alphagov/travel-advice-publisher/blob/master/app/notifiers/publishing_api_notifier.rb

docs/publish_and_notification_workflow.md
```diff
@@ -0,0 +1,20 @@
...
+
+The [jobs order is validated](https://github.com/alphagov/travel-advice-publisher/blob/master/app/notifiers/publishing_api_notifier.rb#L43-43) to ensure that the email notification is the final one, before queuing up the [Sidekiq job](http://sidekiq.org/) that perform the notification if the publishing succeeds.
```
gpeng: should be 'performs' here

docs/publish_and_notification_workflow.md
pmanrubia: The tasks within the job are executed in order with no `explicit` dependency among them; but there is an implicit dependency in the array that is passed to the worker (Job), because any task (within the job) depends on the success of the previous one:  [as an exception is thrown if any job fails](https://github.com/alphagov/travel-advice-publisher/blob/master/app/workers/publishing_api_worker.rb#L14-L16).

```ruby
  def perform(jobs)
    jobs.each do |endpoint, content_id, payload|
      payload = payload.symbolize_keys if payload.is_a?(Hash)

      begin
        if endpoint == "send_alert"
          EmailAlertApiWorker.perform_async(payload)
        else
          api.public_send(endpoint, content_id, payload)
        end
      rescue => e
        raise_helpful_error(e, jobs, endpoint, content_id, payload)
      end
    end
  end
```

So, trying to address your question, the tasks:
- put_content
- put_link
- publish
- send_alert

are guaranteed to run sequentially once the job has started, but the `send_alert` task spins up new Job in Sidekiq to communicate with `email-alert-api`

So, at the very end, we have two asynchronous process within a distributed transactional scope which, to my knowledge, it is not possible to guarantee 100% of the times.

It might be worth mentioning something about how jobs are chained...?

boffbowsh: Ah ok, so this is one job that executes multiple tasks in a single Sidekiq job. It's probably worth clarifying this, "ordered sequence of jobs" makes it sound like multiple Sidekiq jobs.

boffbowsh:
👍 apart from mine and Graham's comments

pmanrubia:
@boffbowsh @gpeng I have updated the documentation with your comments. Thank you.

I have squashed and force pushed in preparation for merge.

PR #139, opened 2016-04-15
Send GOVUK-Request-Id HTTP header downstream.

elliotcm:
The workers run in different threads/processes to
the webserver so HTTP headers are lost.

https://trello.com/c/Eddv2AiS/688-for-migrated-formats-update-for-request-tracing

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ end
 if ENV['API_DEV']
   gem 'gds-api-adapters', :path => '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '~> 27.2.0'
+  gem 'gds-api-adapters', '~> 30.2.0'
 end
 
 gem 'logstasher', '0.4.8'
--- a/app/workers/email_alert_api_worker.rb
+++ b/app/workers/email_alert_api_worker.rb
@@ -1,7 +1,9 @@
 class EmailAlertApiWorker
   include Sidekiq::Worker
 
-  def perform(payload)
+  def perform(payload, params = {})
+    GdsApi::GovukHeaders.set_header(:govuk_request_id, params["request_id"])
+
     api.send_alert(payload) if send_alert?
   rescue => e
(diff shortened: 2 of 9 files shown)


app/workers/email_alert_api_worker.rb
```diff
@@ -1,7 +1,9 @@
 class EmailAlertApiWorker
   include Sidekiq::Worker
 
-  def perform(payload)
+  def perform(payload, params = {})
+    GdsApi::GovukHeaders.set_header(:govuk_request_id, params["request_id"])
```
tuzz: Should we `#fetch` or are there cases where "request_id" might be nil?

spec/requests/request_tracing_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      edition: {
+        title: "New title",
+      },
+    }
+    inbound_headers = {
+      "HTTP_GOVUK_REQUEST_ID" => govuk_request_id,
+    }
+
+    put "/admin/editions/#{edition.id}", params, inbound_headers
+    GdsApi::GovukHeaders.clear_headers # Simulate workers running in a separate thread
```
tuzz: 👍

Gemfile
```diff
@@ -25,7 +25,7 @@ end
 if ENV['API_DEV']
   gem 'gds-api-adapters', :path => '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '~> 27.2.0'
+  gem 'gds-api-adapters', '~> 30.2.0'
```
tuzz: 👍

tuzz:
Looks fine. One tiny question, then :shipit:

app/workers/email_alert_api_worker.rb
tuzz: Following a discussion, this is fine.

elliotcm: c4jt321

PR #140, opened 2016-04-18
Save `govuk_request_id` on publish

gpeng:
The `govuk_request_id` is set per request in the headers. Storing this value when an `Edition` is published will allow the database record to be compared with the front end representation and the alert email for monitoring purposes to ensure the publish and alert succeeded.

[Trello](https://trello.com/c/TQoOXFZ4/31-save-request-id-on-publish-in-tap)

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -109,6 +109,8 @@ class Admin::EditionsController < ApplicationController
         flash[:alert] = @edition.errors.full_messages.join(", ")
       end
 
+      PublishRequest.create(request_id: govuk_request_id, edition_id: @edition.id)
+
       redirect_to admin_country_path(@edition.country_slug), :alert => "#{@edition.title} published."
     else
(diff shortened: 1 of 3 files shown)


boffbowsh:
Couldn't this just be a `published_by_request_id` field on the main `Edition` document?

gpeng:
Yes it could... that's the approach in the PR from @steventux. I kept it separate as it seemed like a separate concern to me, recording the act of publishing. This also kept the changes out of `govuk_content_models` for now and isolated the change while we validate the approach and aren't using it elsewhere.

boffbowsh:
If you want to validate the approach without changing content_models, I believe you can add it to `app/enhancements/edition.rb` as a temporary measure.

gpeng:
The other thing that I didn't mention in my previous comment is that it also gives us a definitive published timestamp which I don't think is reliably present on the `edition`. This will be useful when we're running the checks that are yet to be implemented which could also store their passed/failed state on the `PublishRequest` model.

boffbowsh:
If that's something that is valuable and currently missing, we should make that a first class citizen too

PR #145, opened 2016-04-25
[WIP] Add publication checks

gpeng:
Just for discussion so please don't merge.

The basic flow is that `Runner` accepts an array of `PublishRequest` objects to check and then passes them through each check (currently `EmailAlertCheck` and `ContentStoreCheck`). The check objects will mark the supplied `PublishRequest` as having been successfully checked if the check passes. If not they will do nothing.

When the checks are complete `Runner` will mark the `PublishRequest` as having been checked. This will probably involve incrementing a count on the `PublishRequest` which will allow them to be retried until they haven't passed after _n_ retries at which point they are considered to have failed.

`Result` gets populated by runner and will expose a list of failed `PublishRequest` objects and `#success?` to allow a calling rake task to return the failed publishing attempts if necessary.

None of the state on `PublishRequest` has been implemented as yet.

I'd be interested to hear thoughts on retry intervals and number of attempts. Also whether the use `aws-sdk` as opposed to having a public bucket and just using Net::HTTP... and anything else :)

### DIFF
--- a/app/models/publish_request.rb
+++ b/app/models/publish_request.rb
@@ -4,4 +4,21 @@ class PublishRequest
 
   field :edition_id
   field :request_id
+  field :check_count, type: Integer, default: 0
+  field :succeeded, type: Boolean, default: false
+  field :checks_complete, type: Boolean, default: false
+  field :email_received, type: Boolean, default: false
+  field :frontend_updated, type: Boolean, default: false
+
+  MAX_RETRIES = 3
+
+  def register_check_attempt!
+    self.check_count = check_count + 1
+    self.checks_complete = check_count >= MAX_RETRIES
+    if(self.email_received? && self.frontend_updated?)
+      self.succeeded = true
+      self.checks_complete = true
+    end
+    save!
+  end
 end
--- a/app/models/publish_request.rb
+++ b/app/models/publish_request.rb
@@ -25,4 +25,8 @@ class PublishRequest
   def mark_email_received
     self.email_received = true
   end
+
+  def mark_frontend_updated
+    self.frontend_updated = true
+  end
 end
--- a/app/models/publish_request.rb
+++ b/app/models/publish_request.rb
@@ -21,4 +21,8 @@ class PublishRequest
     end
     save!
   end
+
+  def mark_email_received
+    self.email_received = true
+  end
 end
--- a/app/models/publish_request.rb
+++ b/app/models/publish_request.rb
@@ -13,6 +13,33 @@ class PublishRequest
 
   MAX_RETRIES = 3
 
+  def self.awaiting_check
+    #returns the most recent per country_slug
+    #with checks_complete == false
+    #and created_at more than 5 minutes ago
+    ids = collection.aggregate(
+      [
+        {
+          "$group" => {
+            _id: "$country_slug",
+            maxCreatedAt: {"$max" => "$created_at"},
+            publishRequest: {
+              "$last": "$$ROOT"
+            }
+          }
+        },
+        {
+          "$group" => {
+            _id: "$publishRequest._id"
+          }
+        }
+      ]
+    ).map{|result| result["_id"]}
+    where(:_id.in => ids)
+      .where(checks_complete: false)
+      .where(:created_at.lt => 5.minutes.ago)
+  end
+
   def register_check_attempt!
     self.check_count = check_count + 1
     self.checks_complete = check_count >= MAX_RETRIES
@@ -29,5 +56,10 @@ class PublishRequest
 
   def mark_frontend_updated
     self.frontend_updated = true
+  end
+
+  def mark_superseded
+    # self.superseded = true
+    # self.checks_complete = true
   end
 end
--- a/app/models/publish_request.rb
+++ b/app/models/publish_request.rb
@@ -9,6 +9,7 @@ class PublishRequest
   field :checks_complete, type: Boolean, default: false
   field :email_received, type: Boolean, default: false
   field :frontend_updated, type: Boolean, default: false
+  field :country_slug, type: String
 
   MAX_RETRIES = 3
 
--- a/lib/publication_check/content_store_check.rb
+++ b/lib/publication_check/content_store_check.rb
@@ -1,0 +1,36 @@
+module PublicationCheck
+  class ContentStoreCheck
+    attr_reader :request_id, :publish_request
+
+    def run(publish_request)
+      @publish_request = publish_request
+      @request_id = publish_request.request_id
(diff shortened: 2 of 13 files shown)


lib/publication_check/content_store_check.rb
```diff
@@ -0,0 +1,36 @@
...
+      edition.country_slug
+    end
+
+    def page_response
+      Net::HTTP.get_response(URI(content_store_url))
+    end
+
+    def page_has_request_id?
+      response_json = JSON.parse(page_response.body)
+      content_store_request_id = response_json["publishing_request_id"]
```
steventux: [`publishing_request_id` is currently sent in the details hash](https://www.gov.uk/api/content/foreign-travel-advice/british-antarctic-territory).

gpeng: Thanks... Fixed in e0edb16

steventux:
:+1: this looks great. It makes sense to have a state field on `PublishRequest` so that you can easily scope to failed/unchecked checks. As for number and frequency of retries, it would be good to make this easily configurable while we determine what the noise levels are from this monitoring.

lib/publication_check/content_store_check.rb
```diff
@@ -0,0 +1,36 @@
...
+    def run(publish_request)
+      @publish_request = publish_request
+      @request_id = publish_request.request_id
+      page_has_request_id?
+    end
+
+  private
+
+    def content_store_url
+      "https://www.gov.uk/api/content/foreign-travel-advice/#{country_slug}"
```
pmanrubia: Minor: The `URL` would change per environment?

pmanrubia:
Great work @gpeng . I like the idea to isolate the validations within objects 👍 .

Not sure about the best way to do it, but with the information you are persisting we would also be able to know if an email with outdated information has been sent to the user. 
If `email_received==true` and `front_end_updated == false`, then we might need to take an action.
This would cover scenarios when an issue between the Content Store and the Publishing API prevent a country to be updated before the email gets to the users.

lib/publication_check/content_store_check.rb
gpeng: Ah... good point. I wasn't considering this running in an environment other than production.

gpeng: Fixed in 0735c3d

boffbowsh:
How does this actually get run?

gpeng:
Haven't got that bit yet. Just going to look at it. My intention is to firstly add a scope to `PublishRequest` that will return `checks_complete? == false` that are older than a few minutes (and then only the latest for each `Edition`) and pass them into this setup within a `rake` task that can be called from Jenkins.

app/models/publish_request.rb
```diff
@@ -4,4 +4,29 @@ class PublishRequest
...
   field :request_id
+  field :check_count, type: Integer, default: 0
+  field :succeeded, type: Boolean, default: false
+  field :checks_complete, type: Boolean, default: false
+  field :email_received, type: Boolean, default: false
+  field :frontend_updated, type: Boolean, default: false
+
+  MAX_RETRIES = 3
+
+  def register_check_attempt!
```
boffbowsh: As this is Mongo, we could just add the check timestamp to an array.

app/models/publish_request.rb
```diff
@@ -4,4 +4,29 @@ class PublishRequest
...
+    self.checks_complete = check_count >= MAX_RETRIES
+    if(self.email_received? && self.frontend_updated?)
+      self.succeeded = true
+      self.checks_complete = true
+    end
+    save!
+  end
+
+  def mark_email_received
+    self.email_received = true
```
boffbowsh: Similarly this could be the timestamp it was received at.

lib/publication_check/content_store_check.rb
```diff
@@ -0,0 +1,38 @@
...
+      @request_id = publish_request.request_id
+      result = page_has_request_id?
+      publish_request.mark_frontend_updated if result
+      result
+    end
+
+  private
+
+    def content_store_url
+      "#{Plek.find('www-origin')}/api/content/foreign-travel-advice/#{country_slug}"
```
boffbowsh: This will go through numerous caching and HTTP layers, it might be better to hit content store directly, like a frontend app would.

lib/publication_check/email_alert_check.rb
```diff
@@ -0,0 +1,42 @@
...
+        get_object
+        publish_request.mark_email_received
+        return true
+      rescue Aws::S3::Errors::NoSuchKey
+        return false
+      end
+    end
+
+    def get_object
+      s3.get_object(
```
boffbowsh: I think a simple `HEAD` request with `Net::HTTP` would be better. Adds less complexity, one less gem to load.

lib/publication_check/content_store_check.rb
gpeng: Ideally I think we want it to go through the cacheing etc as we want to ensure that the content is actually being rendered on the frontend and that the cacheing isn't an issue. Is this is far out to 'the edge' as we can get and still be able to run on the `integration` and `staging`?

lib/publication_check/email_alert_check.rb
```diff
@@ -0,0 +1,42 @@
...
+        bucket: bucket,
+        key: object_key
+      )
+    end
+
+    def bucket
+      "govuk-email-alert-notifications"
```
boffbowsh: This should be configurable as an env var

lib/publication_check/runner.rb
```diff
@@ -0,0 +1,20 @@
...
+    DEFAULT_CHECKS = [
+      EmailAlertCheck,
+      ContentStoreCheck
+    ]
+
+    def self.run_check(publish_requests, checks = DEFAULT_CHECKS)
+      Result.new.tap do |result|
+        publish_requests.each do |publish_request|
+          checks.map(&:new).each do |check|
+            check.run(publish_request)
```
boffbowsh: I think `check.new.run(publish_request)` would be clearer than the `map(&:new)` above

lib/publication_check/content_store_check.rb
boffbowsh: True. In which case `Plek.website_root` should do what you want. https://github.com/alphagov/plek/blob/master/lib/plek.rb#L105

gpeng: Ah great. Nice one. Perfect.

gpeng:
Closed in favour of #150 and subsequent PRs to add the other checks from here when the supporting tasks on AWS etc have been completed.

PR #150, opened 2016-06-27
Add page published check

gpeng:
This PR adds a mechanism for checking that a `TravelAdviceEdition` is being rendered on the front end following publication.

This will be run from a Jenkins job via `rake` allowing an alert to be raised in the event that the published edition fails to make it's way to the content store.

The basic workflow is to retrieve the latest `PublishRequest` that has not previously completed the checks successfully and, if it is older than 5 minutes, compare the `request_id` with that rendered into the content store response for the slug. Success or failure is then stored on the `PublishRequest`.

Once a 'run' of checks is complete any that have failed 3 times cause an error message to be returned and the rake task to `exit(2)`. 

[Trello](https://trello.com/c/mjBCZz2N/402-check-publication-has-succeeded-from-within-travel-advice-publisher-medium)

### DIFF
--- a/spec/lib/publication_check/content_store_check_spec.rb
+++ b/spec/lib/publication_check/content_store_check_spec.rb
@@ -45,6 +45,11 @@ module PublicationCheck
         expect(content_store_check.run(publish_request))
           .to be(true)
       end
+
+      it "marks the publish request as frontend updated" do
+        expect(publish_request).to receive(:mark_frontend_updated)
+        content_store_check.run(publish_request)
+      end
     end
 
     context "a response containing a different request id" do
@@ -55,6 +60,11 @@ module PublicationCheck
       it "returns false" do
         expect(content_store_check.run(publish_request))
           .to be(false)
+      end
+
+      it "doesn't mark the publish request as frontend updated" do
+        expect(publish_request).not_to receive(:mark_frontend_updated)
+        content_store_check.run(publish_request)
       end
     end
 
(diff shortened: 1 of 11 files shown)


spec/lib/publication_check/content_store_check_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          }
+        }
+      JSON
+    }
+
+    before do
+      ENV["AUTH_USERNAME"] = "dave"
+      ENV["AUTH_PASSWORD"] = "lemmein"
+      stub_request(:get, "http://dave:lemmein@www.dev.gov.uk/api/content/foreign-travel-advice/andorra").
+        to_return(:status => 200, :body => response_body, :headers => {})
```
tvararu: Minor: inconsistent `.`.

spec/lib/publication_check/content_store_check_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+            "publishing_request_id": "#{response_publishing_request_id}"
+          }
+        }
+      JSON
+    }
+
+    before do
+      ENV["AUTH_USERNAME"] = "dave"
+      ENV["AUTH_PASSWORD"] = "lemmein"
+      stub_request(:get, "http://dave:lemmein@www.dev.gov.uk/api/content/foreign-travel-advice/andorra").
```
tvararu: Looks like you could DRY up the `ENV` allocations and this string but it's not important given they're in such close proximity.

tvararu:
Looks fine to me. 👍  Feel free to ignore the minor pedantic comments.

spec/lib/publication_check/content_store_check_spec.rb
gpeng: Fixed in 7da0a75

spec/lib/publication_check/content_store_check_spec.rb
gpeng: I did try but I think it's probably clearer as is so I've left it.

PR #155, opened 2016-09-12
Change links presenter to send array of content ids

dougdroper:
Passthrough hashes have been removed from the publishing API, sending the
hashes is not valid anymore.


app/presenters/index_links_presenter.rb
```diff
@@ -2,7 +2,10 @@ class IndexLinksPresenter
   def self.present
     {
       :links => {
-        "parent" => BreadcrumbsPresenter.present_for_index
+        "parent" => [
+          "b9849cd6-61a7-42dc-8124-362d2c7d48b0",
```
tijmenb: There can legally* be only one item in `parent`. I guess the parent in this case should be the travel abroad: https://www.gov.uk/api/content/browse/abroad/travel-abroad

tijmenb:
Since we're hardcoding the related links in government-frontend, I guess we don't really need to set the `related` links here.

app/presenters/index_links_presenter.rb
dougdroper: Yes, except that this changes the behaviour slightly: technically multipage-frontend is currently in this structure that it uses as its breadcrumbs, so wanted to keep parity. 

Although since it is hardcoded it won't make a difference, Im happy to just have the one parent, and also drop related links. What do you think?

tijmenb: MPF should be using the same system as the other pages, following the parent recursively.

Perhaps to not confuse developers altogether, drop both `parent` and `related`? Then our team will pick it up when we standardise the breadcrumbs/sidebar across the GOV.UK.

dougdroper:
Ok, updated

tijmenb:
@gpeng merged this to quickly deploy, because publishing-api had been deployed this morning, making this app error when sending the links.

I've updated the title to reflect the new thing this does.

PR #162, opened 2016-11-17
FOR DISCUSSION: Preview historical versions within app

danielroseman:
This is a prototype of how we might move the "preview" of historical versions of travel-advice away from private-frontend.

FCO have confirmed that they don't need the full in-site preview for these; their only use for them is for printing PDFs in response for queries from the public or travel operators/insurance companies about what the advice was on a certain date.

This PR therefore presents the page within the TAP itself, and replaces the link to private-frontend with an internal link for historical versions, or one to draft-origin for the current preview.

To allow this I have copied over a couple of templates, and a small amount of presenter code, from multipage-frontend, which is obviously not ideal. It also requires the addition of Slimmer, which even though it is disabled on all other pages still adds a bit of overhead.

Comments and alternative approaches welcome.

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -57,6 +57,13 @@ class Admin::EditionsController < ApplicationController
     end
   end
 
+  def preview
+    edition = TravelAdviceEdition.find(params[:edition_id])
+    country = Country.find_by_slug(edition.country_slug)
+    @presenter = EditionPreviewPresenter.new(edition, country)
+    render layout: "preview"
+  end
+
   private
   def permitted_edition_attributes
     params[:edition].permit(
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -1,5 +1,8 @@
 class Admin::EditionsController < ApplicationController
+  include Slimmer::Headers
+  include Slimmer::GovukComponents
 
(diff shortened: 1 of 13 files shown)


boffbowsh:
I think this is probably ok as a concept. For reference, here's the PDF of a draft stack document: 

[draft-preview.pdf](https://github.com/alphagov/travel-advice-publisher/files/599579/draft-preview.pdf)
And here's this version: 
[history-preview.pdf](https://github.com/alphagov/travel-advice-publisher/files/599582/history-preview.pdf)

tijmenb:
Perhaps this could be named something like "previous edition" or "historical edition" to differentiate it from previewing something on the draft stack?

app/controllers/admin/editions_controller.rb
```diff
@@ -1,5 +1,8 @@
 class Admin::EditionsController < ApplicationController
```
tijmenb: Perhaps it's clearer to put everything related to historical editions in a separate controller, so so can drop the all the `before_filters`.

danielroseman:
@boffbowsh I don't think that's quite the right comparison - they use the "Print entire guide" link from the bottom of the page, which looks a bit different.

boffbowsh:
Ah yes, you're right. Example here: https://www.gov.uk/foreign-travel-advice/argentina/print

danielroseman:
Closing until we determine what we want to do.

PR #165, opened 2016-12-21
Preview historical versions in app

danielroseman:
Replace the link to private-frontend for 'preview' of historical versions with a locally rendered version using the print template.

FCO have confirmed that they only use the historical preview for creating PDFs to send in response to enquiries about what the advice was on a particular date. Removing the link to private-frontend helps in the process of deprecating that site and disentangling TAP from content-models.

In addition this allows us to use the draft site for previewing the actual current draft, and link to the live site for the published version.

Currently this is protected by a feature flag so that we can try the functionality on integration.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,7 @@ gem 'logstasher', '0.4.8'
 gem 'airbrake', '~> 4.0.0'
 gem 'govspeak', '~> 3.5.2'
 gem 'govuk_sidekiq', '0.0.4'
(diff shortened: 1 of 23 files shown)


tijmenb:
Code looks good, but I have a hunch that calling it "preview" may confuse this with the preview on draft. Perhaps `historical_edition` / `HistorialEditionPresenter`?

danielroseman:
@tijmenb good point, I've changed the names and rebased.

PR #174, opened 2017-03-08
Remove govuk_content_models

thomasleese:
[Trello card](https://trello.com/c/uTYClH73/893-remove-tap-dependency-on-govuk-content-models-2).


app/models/travel_advice_edition.rb
```diff
@@ -0,0 +1,243 @@
...
+    # format that is not very useful for displaying in the flash. Extract them
+    # out in a more readable way.
+    return if errors.delete(:parts).blank?
+    part_errors = parts.map do |part|
+      "#{part.order}: #{part.errors.full_messages.to_sentence}" if part.errors.present?
+    end
+    errors[:part] = part_errors.select(&:present?).sort.to_sentence
+  end
+
+  def self.attaches(*fields)
```
kevindew: This all looks very similar to the attachable module. I'm wondering/hoping - but with low expectations - whether we get this from the `Attachable` module?

app/models/action.rb
```diff
@@ -0,0 +1,66 @@
...
+    REQUEST_AMENDMENTS          = "request_amendments",
+    SEND_FACT_CHECK             = "send_fact_check",
+    RECEIVE_FACT_CHECK          = "receive_fact_check",
+    SKIP_FACT_CHECK             = "skip_fact_check",
+    SCHEDULE_FOR_PUBLISHING     = "schedule_for_publishing",
+    CANCEL_SCHEDULED_PUBLISHING = "cancel_scheduled_publishing",
+    PUBLISH                     = "publish",
+    ARCHIVE                     = "archive",
+    NEW_VERSION                 = "new_version",
+  ]
```
danielroseman: We can remove a bunch of these actions, and much of the code in this model, since TAP doesn't support fact checks or scheduling.

app/models/action.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  NON_STATUS_ACTIONS = [
+    NOTE                    = "note",
+    IMPORTANT_NOTE          = "important_note",
+    IMPORTANT_NOTE_RESOLVED = "important_note_resolved",
+    ASSIGN                  = "assign",
+  ]
```
danielroseman: Again, TAP only supports NOTE here.

app/models/part.rb
```diff
@@ -0,0 +1,26 @@
+require_dependency "safe_html"
+
+class Part
+  include Mongoid::Document
+
+  embedded_in :guide_edition
+  embedded_in :programme_edition
+  embedded_in :business_support_edition
```
danielroseman: We don't have any of these models in TAP.

app/models/parted.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+  end
+
+  def whole_body
+    self.parts.in_order.map { |i| %(\# #{i.title}\n\n#{i.body}) }.join("\n\n")
+  end
+
+private
+
+  def merge_embedded_parts_errors
```
danielroseman: This could do with cleaning up; this method is partly counteracted by the `extract_part_errors` one in the travel advice model itself, now that they are in the same app it should do the right thing in the first place.

app/models/user.rb
```diff
@@ -0,0 +1,78 @@
...
+  def gravatar_url(opts = {})
+    opts.symbolize_keys!
+    "%s.gravatar.com/avatar/%s%s" % [
+      opts[:ssl] ? "https://secure" : "http://www",
+      Digest::MD5.hexdigest(email.downcase),
+      opts[:s] ? "?s=#{CGI.escape(opts[:s])}" : ""
+    ]
+  end
+
+  def progress(edition, action_attributes)
```
danielroseman: Again don't think we need this method, or record_note, assign or unassign, since the actions are not supported.

app/models/travel_advice_edition.rb
```diff
@@ -0,0 +1,264 @@
...
+  end
+
+  def first_version_cant_be_minor_update
+    if self.minor_update && self.version_number == 1
+      errors.add(:minor_update, "can't be set for first version")
+    end
+  end
+
+  def extract_part_errors
+    # govuk_content_models merges in the Parts errors into the main hash in a
```
danielroseman: This comment is no longer relevant and can be deleted.

PR #176, opened 2017-03-08
Add Docker support

thomasleese:


db/seeds.rb
```diff
@@ -1,7 +1,21 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
-#
-# Examples:
-#
-#   cities = City.create([{ name: 'Chicago' }, { name: 'Copenhagen' }])
-#   Mayor.create(name: 'Emanuel', city: cities.first)
+def seed
+  unless Rails.env.development?
```
kevindew: do we need this? just figuring that we probably want to run e2e tests in a rails prod environment and think risks of accidentally running this on our server prod environment are low

db/seeds.rb
```diff
@@ -1,7 +1,21 @@
...
+  gds_organisation_id = "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+
+  user = User.create!(
+    name: "Test user",
+    permissions: ["signin", "gds_editor"],
+    organisation_content_id: gds_organisation_id,
+  )
+end
+
+seed
```
kevindew: shall we just run this as a procedure rather than a method?

config/mongoid.yml
```diff
@@ -1,12 +1,12 @@
...
   clients:
     default:
-      uri: mongodb://localhost/govuk_content_development
+      uri: <%= ENV["MONGODB_URI"] || "mongodb://localhost/govuk_content_development" %>
 
 test:
   clients:
     default:
-      uri: mongodb://localhost/govuk_travel_advice_publisher_test
+      uri: <%= ENV["MONGODB_URI"] || "mongodb://localhost/govuk_travel_advice_publisher_test" %>
```
kevindew: do you mean `TEST_MONGODB_URI` here?

db/seeds.rb
thomasleese: No need for it, I just left it in because it was in the `specialist-publisher` seed.

PR #178, opened 2017-03-13
Change rendering app to government-frontend

mcgoooo:
This is part of the template consolidation work, all functionality has been duplicated in
government frontend and this is the final step

https://github.com/alphagov/government-frontend/pulls?q=is%3Apr+travel+advice+is%3Aclosed

once this has been releases, we will run the rake task to replublish

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ advice](http://www.gov.uk/foreign-travel-advice).  The application persists
 content in MongoDB and in the downstream
 [publishing-api](https://github.com/alphagov/publishing-api).  Travel advice
 content is rendered by
-[multipage-frontend](https://github.com/alphagov/multipage-frontend).
+[government-frontend](https://github.com/alphagov/government-frontend).
 
 ## Dependencies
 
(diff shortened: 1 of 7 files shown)


tijmenb:
Can you change https://github.com/alphagov/travel-advice-publisher/blob/656c3ef85ea0bd84cd799a9e2f8cd926ad6aca26/app/presenters/search_payload_presenter.rb#L21 too? That will update the search index.

fofr:
> Can you change https://github.com/alphagov/travel-advice-publisher/blob/656c3ef85ea0bd84cd799a9e2f8cd926ad6aca26/app/presenters/search_payload_presenter.rb#L21 too? That will update the search index.

What's the affect of that still showing frontend? ie the app before multipage-frontend.

fofr:
Tests are failing because there's a check for the new search change that's not been updated.

mcgoooo:
@tijmenb @fofr tbh, i can't actually run the tests anywhere but here, could one of you [update the documentation](https://github.com/alphagov/travel-advice-publisher#running-the-test-suite) if you have got the tests running

fofr:
@tijmenb The travel advice index page will still be served from frontend. Does the change you requested still need to be made?

mcgoooo:
@fofr i believe this is the correct change, if you follow the code through, it gets to here
https://github.com/alphagov/travel-advice-publisher/blob/66fbe3497b88d177b90e6d54df374c62f37b8eaa/app/models/enhancements/travel_advice_edition.rb#L25-L28

PR #188, opened 2017-04-26
Register routes as exact rather than prefix

thomasleese:
Currently routes are registered as prefixes, this means that gibberish routes such as '/foreign-travel-advice/france/fslajasifj' will end up returning a content item rather than giving a 404. This means that frontend applications will have to include their own logic to determine whether the part exists or not.

[Trello Card](https://trello.com/c/2sdS0Aps/905-stop-using-prefix-routes-for-travel-advice-publisher-2)

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -66,11 +66,22 @@ private
     "/foreign-travel-advice/#{edition.country_slug}"
   end
 
-  def routes
+  def base_routes
     [
-      { "path" => base_path, "type" => "prefix" },
+      { "path" => base_path, "type" => "exact" },
       { "path" => "#{base_path}.atom", "type" => "exact" },
+      { "path" => "#{base_path}/print", "type" => "exact" },
     ]
+  end
+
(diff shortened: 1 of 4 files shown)


kevindew:
Looks good but I think it might be missing the print / email-signup ones. 

eg: https://www.gov.uk/foreign-travel-advice/thailand/email-signup and https://www.gov.uk/foreign-travel-advice/thailand/print from https://www.gov.uk/foreign-travel-advice/thailand

Not really sure how we work out exactly what the definitive list of magic ones are though :-/

thomasleese:
It looks like `https://www.gov.uk/foreign-travel-advice/thailand/email-signup` was already being sent separately anyway, via [edition_presenter.rb](https://github.com/alphagov/travel-advice-publisher/blob/master/app/presenters/email_alert_signup/edition_presenter.rb).

Looking in the routes for `government-frontend`, I think it means we don't need to register the `/print` version as a route in the content item because it is captured there as a variant of the Rails route.

Does that look about right to you as well?

PR #189, opened 2017-04-26
Stop rendering Govspeak

thomasleese:
Instead we rely on the Publishing API to do the rendering for us.

[Trello card](https://trello.com/c/jTMOiB78/921-stop-rendering-govspeak-in-tap)


kevindew:
This looks good 👍  - though before merging this in we should run the `rake govspeak:compare[travel-advice-publisher]` checks to see if TAP is relying on anything we don't support.

thomasleese:
```
Same HTML: 234/459
Trivial Differences: 225/459
Other differences: 0/459
```

kevindew:
Great - that's a lot better than Specialist Publisher was

PR #194, opened 2017-07-11
Send past part slugs to the Publishing API

thomasleese:
This means that past routes can still be stored and redirected appropriately.

[Trello Card](https://trello.com/c/7Dn3P7B3/954-store-past-routes-so-they-can-be-redirected-2)

### DIFF
--- a/app/presenters/edition_presenter.rb
+++ b/app/presenters/edition_presenter.rb
@@ -75,9 +75,25 @@ private
     ]
   end
 
+  def previous_part_slugs
+    TravelAdviceEdition.where(country_slug: edition.country_slug)
+      .pluck("parts.slug")
+      .compact
+      .flat_map { |slugs| slugs.map { |slug| slug["slug"] } }
+      .uniq
+  end
+
+  def current_part_slugs
+    edition.parts.map(&:slug)
+  end
+
+  def all_part_slugs
+    (previous_part_slugs + current_part_slugs).uniq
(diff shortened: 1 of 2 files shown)


app/presenters/edition_presenter.rb
```diff
@@ -75,9 +75,25 @@ def base_routes
     ]
   end
 
+  def previous_part_slugs
+    TravelAdviceEdition.where(country_slug: edition.country_slug)
+      .pluck(:"parts.slug")
```
kevindew: don't think you need to convert this to a symbol

app/presenters/edition_presenter.rb
```diff
@@ -75,9 +75,25 @@ def base_routes
...
+      .compact
+      .flat_map { |slugs| slugs.map { |slug| slug["slug"] } }
```
kevindew: I did (but couldn't think of any) wonder if there was a better way to do this.

app/presenters/edition_presenter.rb
thomasleese: Yep, fixed!

app/presenters/edition_presenter.rb
thomasleese: Same, it's annoying that Mongo gives back plucked nested fields with the full structure since it means I couldn't even get Mongo to to do the `uniq`.

PR #195, opened 2017-07-19
Add publishing e2e tests stage to Jenkinsfile

emmabeynon:
https://trello.com/c/NwvQLJMs/957-run-e2e-tests-as-part-of-content-store-and-travel-advice-publisher-builds-1


emmabeynon:
> Not part of this PR, but I wonder if we can get this app to use govuk.buildProject()

Yes, would be nice to use that instead of all this code.  Will write up a story for the back log 👍

PR #196, opened 2017-07-26
Send update type with put content

emmabeynon:
We are deprecating sending the update_type in publish requests made to the Publishing API, and now expect clients to send update_type in PUT requests instead. We also upgrade gds-api-adapters to bring this app up to date with the new expected behaviour.

https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -9,7 +9,7 @@ namespace :publishing_api do
 
     api_v2.put_content(presenter.content_id, presenter.render_for_publishing_api)
     api_v2.patch_links(TravelAdvicePublisher::INDEX_CONTENT_ID, IndexLinksPresenter.present)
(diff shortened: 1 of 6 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -9,7 +9,7 @@ namespace :publishing_api do
 
     api_v2.put_content(presenter.content_id, presenter.render_for_publishing_api)
```
thomasleese: Do we know that `render_for_publishing_api` includes the `update_type`?

emmabeynon: Yep it does https://github.com/alphagov/travel-advice-publisher/blob/a97d0ea7536657315eceb717869fe9e83fc980e3/app/presenters/index_presenter.rb#L26

PR #197, opened 2017-07-27
Update responses to behave like a hash

emmabeynon:
Following on from the gds-api-adapters upgrade to 47.2, OpenStruct responses are now longer supported and must now be hashes.

### DIFF
--- a/app/models/travel_advice_edition.rb
+++ b/app/models/travel_advice_edition.rb
@@ -247,7 +247,7 @@ private
       define_method("upload_#{field}") do
         begin
           response = TravelAdvicePublisher.asset_api.create_asset(file: instance_variable_get("@#{field}_file"))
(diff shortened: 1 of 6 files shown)


lib/tasks/rummager.rake
```diff
@@ -2,14 +2,14 @@ namespace :rummager do
   desc "Indexes the main travel advice page in Rummager"
   task index: :environment do
     slug = "foreign-travel-advice"
-    record = OpenStruct.new(
+    record = {
```
thomasleese: I think this needs to remain as an `OpenStruct`, it looks like the `RummagerNotifier` passes it on to `SearchIndexer` which passes it to `SearchPayloadPresenter.call` which expects to see an openstruct-like object.

PR #200, opened 2017-07-27
Hash responses from gds-api-adapters as strings not symbols

thomasleese:
The responses come through as a hash of strings, not symbols,
so we need to index them in that way.


thomasleese:
E2E tests also passing: https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/test-against/118

PR #206, opened 2017-10-18
Use govuk buildproject in jenkinsfile

steventux:
https://trello.com/c/BQCjFGpq/3-investigate-travel-advice-publisher-using-buildproject-groovy-script

This PR simplifies Travel Advice's Jenkinsfile to use `govuk.buildProject`.
In order to do this some dependencies were updated to satisfy the `gov-lint-sass` build step.
I've tested the sass changes on integration and can't see any difference with staging.

PR #207, opened 2017-10-18
Use govuk buildproject in jenkinsfile

steventux:
https://trello.com/c/BQCjFGpq/3-investigate-travel-advice-publisher-using-buildproject-groovy-script

This PR simplifies Travel Advice's Jenkinsfile to use govuk.buildProject.
In order to do this some dependencies were updated to satisfy the gov-lint-sass build step.
I've tested the sass changes on integration and can't see any difference with staging.

There's a minor version downgrade of `sass` here, necessary for the `govuk-lint` gem because this depends on `sass_lint` which specifies `sass ~> 3.4.20`. After some discussion with @fofr I've opted for linting rather than disabling it in the build.

This is a follow up from https://github.com/alphagov/travel-advice-publisher/pull/206 which I can't reopen.


app/assets/stylesheets/scaffold.scss
```diff
@@ -1,16 +1,19 @@
...
 // Override the font-size for main textareas
+// scss-lint:disable QualifyingElement
 .formtastic textarea.form-control {
   font-size: 17px;
   line-height: 26px;
-  color: #555;
+  color: $medium-dark-grey;
 }
 
+// scss-lint:disable QualifyingElement
```
fofr: I think you need to re-enable this again afterwards.

steventux:
@fofr updated, I've re-enabled at the appropriate point.

PR #208, opened 2017-10-19
Conditionally draft newly created editions to the Publishing API

steventux:
https://trello.com/c/vU0gCJds/2-stop-sending-empty-drafts-to-publishing-api

When a country has no previous editions a newly created edition will have no content and will not pass schema validation in the Publishing API.
The test setup in the e2e tests has highlighted this problem which we don't see in real life as we have yet to introduce a new country via the Publishing API.
It's still an edge case we should close off for the [sake of the e2e tests](https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/TAP_e2e_tests/1/artifact/tmp/errors-verbose.log).

cc @kevindew

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -16,9 +16,11 @@ class Admin::EditionsController < ApplicationController
     end
 
     if edition.save
-      notifier.put_content(edition)
-      notifier.patch_links(edition)
-      notifier.enqueue
+      unless edition.first_draft
+        notifier.put_content(edition)
+        notifier.patch_links(edition)
+        notifier.enqueue
+      end
       redirect_to edit_admin_edition_path(edition)
     else
       redirect_to admin_country_path(@country.slug), alert: "Failed to create new edition"
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -22,7 +22,7 @@ class Country
     elsif (latest_edition = editions.first)
       latest_edition.build_clone
     else
-      TravelAdviceEdition.new(country_slug: slug, title: "#{name} travel advice")
+      TravelAdviceEdition.new(country_slug: slug, title: "#{name} travel advice", first_draft: true)
     end
   end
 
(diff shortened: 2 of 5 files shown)


kevindew:
Argh sorry for the delay Steve 😞 Thanks for having a look into it

Looks like there's still a bunch of errors in the log though of the same type, so not sure this has resolved it? I also wonder if there's a way we can do it without passing first_draft around. 

I'll have a quick look.

steventux:
Oh right, perhaps I was looking at the wrong log output but [this log](https://ci.integration.publishing.service.gov.uk/job/publishing-e2e-tests/job/TAP_e2e_tests/1/artifact/tmp/errors-verbose.log) looked to be a failure where some params were populated by the presented but crucially no edition body or other cloned bits of content. Do you see similar failures? My thinking was that because the E2E tests don't setup a previous edition the edition that is created goes via this condition: https://github.com/alphagov/travel-advice-publisher/blob/bd886ec2284570609bb3573f580fa3aec1155a0a/app/models/country.rb#L25 and doesn't clone any content, so the presenter presents an empty and invalid edition.
Or perhaps the transient `first_draft` property isn't working as expected?

kevindew:
Yeah I'm not sure if it's just not working properly or not 😕 

The error we always seem to get is "The property '#/details/summary/0/content' of type NilClass did not match the following type: string" where I wonder if we can just initialise with that as an empty string rather than nil. But I found myself at a total loss from where the countries actually appear from, so just in the process of booting up E2E tests to see what I can gleam

kevindew:
I think [this](https://github.com/alphagov/travel-advice-publisher/commit/03e929d9bf9b9ee148f62dbd64138ce076342202) might be enough to sort the issue - just gave it a whirl and saw no errors logged.

steventux:
This doesn't stop sending an empty draft to the publishing API, but then given how infrequently that ever happens I'm happy if you are.

kevindew:
Yeah, though it looks like Publishing API can handle the bare minimum that TAP sends through (which I didn't realise previously) so does mean you don't have to worry about having a conditional as to whether something is previewable or not.

kevindew:
I'll open up a PR

steventux:
:+1 thanks @kevindew I'll close this.

PR #212, opened 2017-11-27
Update rummager format to `travel-advice`

dwhenry:
This only affects country pages, not the main index page
which has been left using `custom-application`.

This change has been made to simplify the migration of
travel advice documents to being index via the publishing API.

https://trello.com/c/36ZgSxPD/465-we-are-indexing-travel-advice-index-from-the-publishing-api

### DIFF
--- a/spec/tasks/rummager_rake_spec.rb
+++ b/spec/tasks/rummager_rake_spec.rb
@@ -1,0 +1,34 @@
+require 'spec_helper'
+require 'rake'
+require 'gds_api/test_helpers/rummager'
+
+describe "rummager rake tasks", type: :rake_task do
+  include GdsApi::TestHelpers::Rummager
+
+  before do
+    Rake.application = nil # Reset any previously loaded tasks
+    Rails.application.load_tasks
+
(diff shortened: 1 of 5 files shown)


spec/tasks/rummager_rake_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    Rake.application = nil # Reset any previously loaded tasks
+    Rails.application.load_tasks
+
+    stub_any_rummager_post
+  end
+
+  describe "rummager:index" do
+    let(:task) { Rake::Task['rummager:index'] }
+
+    it 'registors the travel advice index page in rummager' do
```
suzannehamilton: Typo: registors

dwhenry:
@suzannehamilton I got a reply from Nick saying it should be fine, but nothing from Tijman, he isn't online. I tested it on integration and it all looks fine. I will be putting it live on staging and then might ask Mark if he can test it to ensure it is working as this is travel advise and I would hate to be the one to break it.. :)

PR #226, opened 2017-12-07
Upgrade to Rails 5.1.4

danielburnley:
https://trello.com/c/40saJAyG/49-rails-upgrade-for-travel-advice-publisher


config/initializers/new_framework_defaults_5_1.rb
```diff
@@ -0,0 +1,14 @@
+# Be sure to restart your server when you modify this file.
```
tijmenb: I haven't seen this file before, can we upgrade so we don't need it?

danielburnley: Yep - I can paste the default into the main config

PR #254, opened 2017-12-21
Add Link and LinkCheckReport models

camillavk:
Add the Link and LinkCheckReport models. These will be used to store the status of links on a TravelAdviceEdition when checked against the LinkChecker API. This PR will be followed by:
- New controller for creating/sending batch requests to the API
- New callback controller to update the LinkCheckReport and Links
- Implementation of UI to kick-off poll check and show results

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -1,0 +1,17 @@
+class Link
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  embedded_in :link_check_report
+
+  field :uri, type: String
+  field :status, type: String
+  field :checked_at, type: DateTime
+  field :check_warnings, type: Array
+  field :check_errors, type: Array
+  field :problem_summary, type: String
+  field :suggested_fix, type: String
+
+  validates :uri, presence: true
+  validates :status, presence: true
+end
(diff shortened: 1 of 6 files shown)


app/models/link.rb
```diff
@@ -0,0 +1,17 @@
+class Link
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  embedded_in :link_check_report
+
+  field :uri, type: String
+  field :status, type: String
+  field :checked, type: DateTime
```
thomasleese: I would call this `checked_at` or `checked_date`.

app/models/link_check_report.rb
```diff
@@ -0,0 +1,18 @@
...
+  include Mongoid::Document
+  include Mongoid::Timestamps
+
+  embeds_many :links
+
+  accepts_nested_attributes_for :links
+
+  field :batch_id, type: Integer
+  field :status, type: String
+  field :edition_id, type: String
```
thomasleese: Instead of doing this, can we just do `embedded_in :travel_advice_edition`?

app/models/link.rb
camillavk: I've gone with `checked_at` here

PR #259, opened 2018-01-02
Add Link Checker API Callback endpoint

SebAshton:
This PR adds an endpoint for the `link-checker-api` to POST the results of batch checks to.

Depends on https://github.com/alphagov/govuk-puppet/pull/7007

### DIFF
--- a/app/controllers/link_checker_api_controller.rb
+++ b/app/controllers/link_checker_api_controller.rb
@@ -1,0 +1,51 @@
+class LinkCheckerApiController < ApplicationController
+  skip_before_action :verify_authenticity_token
+  skip_before_action :authenticate_user!
+  skip_before_action :require_signin_permission!
+  skip_before_action :set_authenticated_user_header
+  before_action :verify_signature
+
+  rescue_from Mongoid::Errors::DocumentNotFound, with: :render_no_content
+
+  def callback
+    if link_check_report
+      LinkCheckReportUpdater.new(
+        report: link_check_report,
+        payload: params
+      ).call
+    end
+
+    render_no_content
+  end
+
+private
+
+  def render_no_content
+    head :no_content
+  end
+
+  def batch_id
+    params.require(:id)
+  end
+
+  def travel_advice_edition
+    @travel_advice_edition ||= TravelAdviceEdition.find_by("link_check_reports.batch_id": batch_id)
+  end
+
+  def link_check_report
+    @link_check_report ||= travel_advice_edition.link_check_reports.find_by(batch_id: batch_id)
+  end
+
+  def verify_signature
+    return unless webhook_secret_token
+    given_signature = request.headers["X-LinkCheckerApi-Signature"]
+    return head :bad_request unless given_signature
+    body = request.raw_post
+    signature = OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new("sha1"), webhook_secret_token, body)
+    head :bad_request unless Rack::Utils.secure_compare(signature, given_signature)
+  end
+
+  def webhook_secret_token
+    Rails.application.secrets.link_checker_api_secret_token
+  end
+end
--- a/app/services/link_check_report_updater.rb
+++ b/app/services/link_check_report_updater.rb
@@ -1,10 +1,4 @@
 class LinkCheckReportUpdater
-  class InvalidReport < RuntimeError
-    def initialize(original_error)
-      @message = original_error.message
-    end
-  end
-
   def initialize(report:, payload:)
     @report = report
     @payload = payload
@@ -14,8 +8,6 @@ class LinkCheckReportUpdater
     update_report!
     links = payload.fetch("links", [])
     update_links!(links)
-  rescue Mongoid::Errors::Validations => e
-    raise InvalidReport.new(e)
   end
 
 private
--- a/app/services/link_check_report_updater.rb
+++ b/app/services/link_check_report_updater.rb
@@ -1,0 +1,53 @@
+class LinkCheckReportUpdater
+  class InvalidReport < RuntimeError
+    def initialize(original_error)
+      @message = original_error.message
+    end
+  end
+
+  def initialize(report:, payload:)
+    @report = report
+    @payload = payload
+  end
+
+  def call
+    update_report!
+    links = payload.fetch("links", [])
+    update_links!(links)
+  rescue Mongoid::Errors::Validations => e
+    raise InvalidReport.new(e)
+  end
+
+private
+
+  attr_reader :report, :payload
+
+  def update_report!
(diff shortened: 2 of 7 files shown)


spec/controllers/link_checker_api_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+RSpec.describe LinkCheckerApiController, type: :controller do
+  include GdsApi::TestHelpers::LinkCheckerApi
+
+  def generate_signature(body, key)
+    OpenSSL::HMAC.hexdigest(OpenSSL::Digest.new("sha1"), key, body)
+  end
+
+  let(:service) { double(:service, call: nil) }
+  let(:link_check_report) {
```
lukemorton: Do block?

spec/controllers/link_checker_api_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    FactoryGirl.create(:travel_advice_edition_with_pending_link_checks,
+                       batch_id: 5,
+                       link_uris: ['http://www.example.com', 'http://www.gov.com']).link_check_reports.first
+  }
```
lukemorton: Could add new line between blocks.

spec/services/link_check_report_updater_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:payload) do
+    {
+      status: 'complete',
+      completed_at: completed_at,
+      links: links_payload
+    }.with_indifferent_access
+  end
+
+  let(:links_payload) do
+    [{
```
lukemorton: Is this the correct style?

spec/controllers/link_checker_api_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  it "POST :update updates LinkCheckReport" do
+    headers = {
+      "Content-Type": "application/json",
+      "X-LinkCheckerApi-Signature": generate_signature(post_body.to_json, Rails.application.secrets.link_checker_api_secret_token)
+    }
+
+    request.headers.merge! headers
+    post :callback, params: post_body
```
thomasleese: I believe you can pass in the headers as an argument here: `post params: params, headers: headers`.

spec/controllers/link_checker_api_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    link_checker_api_batch_report_hash(
+      id: 5,
+      links: [
+        { uri: 'https://www.gov.uk', status: "ok" },
+      ]
+    )
+  end
+
+  before do
+    allow(LinkCheckReport).to receive(:find_by).with(batch_id: 5).and_return(link_check_report)
```
thomasleese: Do we have to do this still sort of thing in travel advice or can we create the objects as real objects?

app/services/link_check_report_updater.rb
```diff
@@ -0,0 +1,53 @@
...
+    @report = report
+    @payload = payload
+  end
+
+  def call
+    update_report!
+    links = payload.fetch("links", [])
+    update_links!(links)
+  rescue Mongoid::Errors::Validations => e
+    raise InvalidReport.new(e)
```
thomasleese: Is this used anywhere? I'm not sure it's useful, and also it won't raised for any `KeyError`s.

spec/controllers/link_checker_api_spec.rb
SebAshton: Unfortunately as this is a controller spec and not a request spec you have to explicitly set the headers

spec/factories.rb
```diff
@@ -36,4 +36,51 @@
...
+        link_uris.map { |uri| FactoryGirl.build(:link, :pending, uri: uri) }
+      end
+    end
+  end
+
+  factory :link do
+    uri "http://www.example.com"
+    status "ok"
+
+    trait :pending do
```
thomasleese: Should this trait be called `:broken`?

spec/services/link_check_report_updater_spec.rb
SebAshton: I can't find any examples to confirm or deny they styling, but RuboCop is ok, so I assume it to be correct.

lukemorton: Ah okay cool.

spec/controllers/link_checker_api_spec.rb
lukemorton: :(

spec/controllers/link_checker_api_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    link_checker_api_batch_report_hash(
+      id: 5,
+      links: [
+        { uri: 'https://www.gov.uk', status: "ok" },
+      ]
+    )
+  end
+
+  before do
+    link_check_report
```
thomasleese: You can use `let!` to make sure the variable is defined always.

spec/factories.rb
SebAshton: Opted to change the status to `pending` and remove the redundant broken trait

app/controllers/link_checker_api_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+        params: params
+      ).call
+    end
+
+    head :no_content
+  end
+
+private
+
+  def find_report_in_travel_advice_edition
```
thomasleese: Why do we have to run the query like this?

thomasleese: I'm guessing this is necessary because it's an embedded document, if we have to query on the edition first, I wonder if we could we have something like which might be a bit nicer:

```ruby
def batch_id
  params.require(:id)
end

def travel_advice_edition
  @travel_advice_edition ||= TravelAdviceEdition.find_by(link_check_reports: { batch_id: batch_id })
end

def link_check_report
  @link_check_report ||= travel_advice_edition.link_check_reports.find_by(batch_id: batch_id)
end
```

SebAshton:
Made those updates @thomasleese

spec/controllers/link_checker_api_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    )
+  end
+
+  context 'when the report exists' do
+    before do
+      expect(TravelAdviceEdition).to receive(:find_by).with("link_check_reports.batch_id": 5).and_call_original
+      expect(LinkCheckReportUpdater).to receive(:new).and_return(service)
+    end
+
+    it "POST :update updates LinkCheckReport" do
```
thomasleese: Instead of testing that we the methods here get called and stubbing out the `service`, could we just check the model has been updated? That feels like a more thorough test to me.

PR #262, opened 2018-01-03
Link check report controller

camillavk:
This PR adds the link checker API and uses it to make `create_batch` requests for Travel Advice Editions. It creates LinkCheckReports with the response from the API.

### DIFF
--- a/app/services/link_check_report_creator.rb
+++ b/app/services/link_check_report_creator.rb
@@ -1,0 +1,63 @@
+require "gds_api/link_checker_api"
+require "govspeak/link_extractor"
+
+class LinkCheckReportCreator
+  include Rails.application.routes.url_helpers
+
+  CALLBACK_HOST = Plek.find("travel-advice-publisher")
+
+  def initialize(travel_advice_edition_id:)
+    @travel_advice_edition_id = travel_advice_edition_id
+  end
+
+  def call
+    link_report = call_link_checker_api
+
+    report = travel_advice_edition.link_check_reports.new(
+      batch_id: link_report.fetch(:id),
+      completed_at: link_report.fetch(:completed_at),
+      status: link_report.fetch(:status),
+      links: link_report.fetch(:links).map { |link| map_link_attrs(link) }
+    )
+
+    report.save!
+  end
+
+private
+
+  attr_reader :travel_advice_edition_id
+
+  def travel_advice_edition
+    @travel_advice_edition ||= TravelAdviceEdition.find(travel_advice_edition_id)
+  end
+
+  def call_link_checker_api
+    callback = link_checker_api_callback_url(host: CALLBACK_HOST)
+
+    TravelAdvicePublisher.link_checker_api.create_batch(
+      uris,
+      webhook_uri: callback,
+      webhook_secret_token: Rails.application.secrets.link_checker_api_secret_token
+    )
+  end
+
+  def govspeak_document
+    Govspeak::Document.new(travel_advice_edition.summary)
+  end
+
+  def uris
+    govspeak_document.extracted_links
+  end
+
+  def map_link_attrs(link)
+    {
+      uri: link.fetch(:uri),
+      status: link.fetch(:status),
+      checked_at: link.fetch(:checked),
+      check_warnings: link.fetch(:warnings, []),
+      check_errors: link.fetch(:errors, []),
+      problem_summary: link.fetch(:problem_summary),
+      suggested_fix: link.fetch(:suggested_fix)
+    }
+  end
+end
--- a/config/initializers/link_checker_api.rb
+++ b/config/initializers/link_checker_api.rb
@@ -1,0 +1,6 @@
+require "gds_api/link_checker_api"
+require "plek"
+
+TravelAdvicePublisher.link_checker_api = GdsApi::LinkCheckerApi.new(
+  Plek.find("link-checker-api")
+)
--- a/spec/controllers/link_check_reports_spec.rb
+++ b/spec/controllers/link_check_reports_spec.rb
@@ -1,0 +1,49 @@
+require "spec_helper"
+
+describe LinkCheckReportsController, type: :controller do
+  describe "#create" do
+    let(:travel_advice_edition) do
+      FactoryGirl.create(:travel_advice_edition,
+                         summary: "[link](http://www.example.com)[link_two](http://www.gov.com)")
+    end
+
+    let(:completed_at) { Time.now }
+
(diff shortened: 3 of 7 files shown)


app/services/link_check_report_creator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    report.save!
+  end
+
+private
+
+  attr_reader :travel_advice_edition_id
+
+  def travel_advice_edition
+    @travel_advice_edition ||= TravelAdviceEdition.find(id: travel_advice_edition_id)
```
thomasleese: Do you need the `id: ` if you use `find`?

config/initializers/link_checker_api.rb
```diff
@@ -0,0 +1,6 @@
+require "gds_api/link_checker_api"
+require "plek"
+
+TravelAdvicePublisher.link_checker_api = GdsApi::LinkCheckerApi.new(
+  Plek.current.find("link-checker-api")
```
thomasleese: I think you can just do `Plek.find` here.

lib/govspeak/link_extractor.rb
```diff
@@ -0,0 +1,24 @@
+module Govspeak
```
thomasleese: I'm wondering if this class (since we've had to put it in many apps) should live in https://github.com/alphagov/govspeak/tree/master/lib/govspeak ?

spec/controllers/link_check_reports_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+            errors: [],
+            problem_summary: "",
+            suggested_fix: ""
+          }
+        ]
+      }
+    end
+
+    before do
+      allow(TravelAdvicePublisher.link_checker_api).to receive(:create_batch).and_return(link_checker_api_response)
```
thomasleese: Instead of doing this, I wonder if we should use `webmock` and the available test helpers here: https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/test_helpers/link_checker_api.rb#L49

spec/services/link_check_report_creator_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  let(:link_check_report) do
+    FactoryGirl.create(:travel_advice_edition_with_pending_link_checks,
+                       batch_id: 1,
+                       link_uris: ['http://www.example.com', 'http://www.gov.com']).link_check_reports.first
+  end
+
+  before do
+    allow(TravelAdvicePublisher.link_checker_api).to receive(:create_batch).and_return(link_checker_api_response)
+    allow(LinkCheckReport).to receive(:new).and_return(link_check_report)
```
thomasleese: Do we need this if we have the object with `FactoryGirl.create`?

spec/controllers/link_check_reports_spec.rb
camillavk: Unfortunately we can't at the moment because the helper doesn't correspond completely with the response from the API. The helper simply returns the list of links whereas I need to have things like the `problem_summary` etc in the response.

lib/govspeak/link_extractor.rb
camillavk: Yup, this probably wouldn't be a bad idea. As the work in this PR will be reliant on that, would it make sense to add it in there as part of the next integration then remove it from here (and from manuals-publisher) once it is deployed?

config/initializers/link_checker_api.rb
camillavk: 👍

spec/services/link_check_report_creator_spec.rb
camillavk: I think because of the fact the `LinkCheckReport` is nested in the `TravelAdviceEdition` the tests do not work without this

camillavk:
@thomasleese I've addressed your comments so this should be ready for re-review now :)

spec/services/link_check_report_creator_spec.rb
thomasleese: Hmm, do you know what the error is? I'm wondering if this indicates a problem with the code rather than the tests.

spec/controllers/link_check_reports_spec.rb
thomasleese: I think it would be better to fix the test helpers so they correctly match what the API responds with as that will help when writing tests in other apps that use these helpers.

lib/govspeak/link_extractor.rb
thomasleese: I think since it's a pretty small change to the `govspeak` gem we could get it turned around quick enough for this PR.

spec/services/link_check_report_creator_spec.rb
camillavk: `undefined method `__metadata' for nil:NilClass`

lib/govspeak/link_extractor.rb
camillavk: 👍 I'll get a PR in to the `govspeak` gem then

camillavk:
This isn't ready for re-review yet, I still need to make the changes to the `govspeak` gem, however I have found a way to use the gds-api test_helpers without making changes to that repo.

camillavk:
@thomasleese I've moved the code around into the relevant commits so this should be all good to go now

app/services/link_check_report_creator.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    TravelAdvicePublisher.link_checker_api.create_batch(
+      uris,
+      webhook_uri: callback,
+      webhook_secret_token: Rails.application.secrets.link_checker_api_secret_token
+    )
+  end
+
+  def uris
+    Govspeak::LinkExtractor.new(Govspeak::Document.new(travel_advice_edition.summary)).call
```
thomasleese: Might be nice to have here:

```ruby
def govspeak_document
  Govspeak::Document.new(travel_advice_edition.summary)
end

def uris
  govspeak_document.extract_links
end
```

app/services/link_check_report_creator.rb
```diff
@@ -0,0 +1,59 @@
+require "gds_api/link_checker_api"
+require "govspeak/link_extractor"
+
+class LinkCheckReportCreator
+  include Rails.application.routes.url_helpers
+
+  CALLBACK_HOST = Plek.find('travel-advice-publisher')
```
thomasleese: Can we use double quotes here.

camillavk:
@thomasleese I've addressed your two comments on this PR

PR #270, opened 2018-01-09
Add link checker ui

camillavk:
This PR adds the UI for the Link Checker onto the editions#edit view. This also displays the relevant information about links that have returned with a broken or caution status on the same page. 

screen shot 2018-01-09 at 15 32 34

### DIFF
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -1,7 +1,4 @@
 class LinkCheckReportsController < ApplicationController
-  skip_before_action :authenticate_user!
-  skip_before_action :require_signin_permission!
-
   def create
     service = LinkCheckReportCreator.new(
       travel_advice_edition_id: link_reportable_params[:travel_advice_edition_id]
--- a/app/controllers/link_check_reports_controller.rb
+++ b/app/controllers/link_check_reports_controller.rb
@@ -1,22 +1,22 @@
 class LinkCheckReportsController < ApplicationController
-  before_action :find_reportable
+  before_action :find_edition
 
   def create
     service = LinkCheckReportCreator.new(
-      travel_advice_edition_id: @reportable
+      travel_advice_edition_id: @edition.id
     )
 
     @report = service.call
 
     respond_to do |format|
-      format.js { render 'link_check_reports/create' }
-      format.html { redirect_to edit_admin_edition_url(@reportable.id) }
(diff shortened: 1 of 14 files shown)


app/controllers/link_check_reports_controller.rb
```diff
@@ -1,20 +1,22 @@
 class LinkCheckReportsController < ApplicationController
-  skip_before_action :authenticate_user!
-  skip_before_action :require_signin_permission!
+  before_action :find_reportable
 
   def create
     service = LinkCheckReportCreator.new(
-      travel_advice_edition_id: link_reportable_params[:travel_advice_edition_id]
+      travel_advice_edition_id: @reportable
```
thomasleese: Should we rename this field to `travel_advice_edition` or perhaps even just `edition`?

app/controllers/link_check_reports_controller.rb
```diff
@@ -1,20 +1,22 @@
...
+      format.js { render 'admin/link_check_reports/create' }
+      format.html { redirect_to edit_admin_edition_url(@reportable.id) }
+    end
   end
 
 private
 
-  def link_reportable_params
-    params.require(:link_reportable).permit(:travel_advice_edition_id)
+  def find_reportable
```
thomasleese: We can probably just call this `find_edition` since I believe travel advice only has one kind of edition.

app/models/travel_advice_edition.rb
```diff
@@ -139,6 +139,12 @@ def whole_body
     self.parts.in_order.map { |i| %(\# #{i.title}\n\n#{i.body}) }.join("\n\n")
   end
 
+  def latest_link_check_report
+    return nil if link_check_reports.empty?
```
thomasleese: I don't think we need this since `.last` returns `nil` if the array is empty.

app/views/admin/editions/edit.html.erb
```diff
@@ -94,6 +94,10 @@
           <% end %>
         </div>
 
+        <div class="col-md-4 link-check-report">
+          <%= render "admin/link_check_reports/link_check_report", reportable: @edition, report: @edition.latest_link_check_report %>
```
thomasleese: Again I wonder if we should just stick with the term `edition` in travel advice publisher.

camillavk:
@thomasleese I've renamed `@reportable` to `@edition` in all the relevant files

PR #271, opened 2018-01-10
Add show endpoint for link check reports

camillavk:
This PR adds the ability to get the status of a link check report based using its id. It provides both a JS and an HTML response for the show endpoint.


app/controllers/link_check_reports_controller.rb
```diff
@@ -14,9 +14,24 @@ def create
...
 private
 
+  attr_reader :link_check_report
+
   def find_edition
     @edition = TravelAdviceEdition.find(params[:edition_id])
   end
+
+  def link_check_report
+    @link_check_report ||= @edition.link_check_reports.find_by(id: params[:id])
```
thomasleese: Should we definitely be caching this? Or could it mean some times this returns out of date data?

camillavk: Ha, I just thought of the same thing so I'm in the process of changing it :D

PR #291, opened 2018-02-07
Publish email signup when edition first published

dwhenry:
Automatically publish the signup page when the country gets it's first edition

https://trello.com/c/oom1LAQE/52-create-new-travel-advice-page-for-cook-islands-tokelau-and-niue

### DIFF
--- a/app/controllers/admin/editions_controller.rb
+++ b/app/controllers/admin/editions_controller.rb
@@ -108,6 +108,7 @@ private
     if @edition.update_attributes(permitted_edition_attributes) && @edition.publish_as(current_user)
       notifier.put_content(@edition)
       notifier.patch_links(@edition)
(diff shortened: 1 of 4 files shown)


dwhenry:
@h-lame understand the risk you point out, I'm not sure that moving this to the controller so that the signup page - which should be static - if republished each time the advice page is updated is the best solution to this.

I feel that adding a test to ensure this path is checked for again any future code change is a better solution. Thoughts?

h-lame:
I think we should explore this a little bit, but if it turns out there's no easy way to check if an edition is the first edition and we should publish the email-signup during create (or maybe we should be doing it via update when the action is "Save and publish"?), then I think this rake task is the thing we should do for now and we can revisit the email-signup thing later.

dwhenry:
@h-lame updated as discussed

PR #292, opened 2018-02-09
Use the external_url_for method where appropriate

cbaines:


cbaines:
(Note to self, Travel Advice Publisher has some very flaky tests, it took 4 builds of this branch for all the tests to pass simultaneously...)

PR #294, opened 2018-02-12
Various dependency updates

kevindew:
The intention here was to update to govuk_app_config and bump the other gems bit there was a reasonable amount of work to do so this has ended up somewhat sizable.

Highlights are probably:

- More consistent Gemfile versioning
- Everything bundle updated
- Drop Timecop dependency for ActiveSupport::Testing::TimeHelpers
- Fix the gds-sso deprecation warning
- Fix FactoryGirl deprecations
- Linting up to date
- App can run all through a Procfile.


kevindew:
@thomasleese that's probably a good idea

PR #334, opened 2018-05-16
Set primary org

chao-xian:
The content types for the Travel Advice Publisher do not set a primary organisation. This PR adds the hard coded value for every item of this content type to be set on publish. We are specifying that HMRC as the primary org for these.

This is part of wider work to set a primary org on all content types that don't currently have this value set.


chao-xian:
[(3) Set Primary Org for Travel Advice](https://trello.com/c/V8dTpa6u/329-3-set-primary-org-for-travel-advice)

config/application.rb
```diff
@@ -22,12 +22,16 @@ module TravelAdvicePublisher
...
   INDEX_CONTENT_ID = "08d48cdd-6b50-43ff-a53b-beab47f4aab0".freeze
+  # The parent page is /browse/abroad/travel-abroad
+  INDEX_PARENT_CONTENT_ID = "b9849cd6-61a7-42dc-8124-362d2c7d48b0".freeze
   INDEX_EMAIL_SIGNUP_CONTENT_ID = "1aebfc97-7723-4cb6-82f4-434639efc185".freeze
 
   COUNTRY_FORMAT = "travel_advice".freeze
 
   INDEX_FORMAT = "travel_advice_index".freeze
 
+  PRIMARY_ORG_CONTENT_ID = "6667cce2-e809-4e21-ae09-cb0bdc1ddda3".freeze
```
thomasleese: Should this be FCO rather than HMRC?

MatMoore: Yep, I mixed them up.

tijmenb:
(Note that I've updated the PR description to say that the publishing org is FCO, not HMRC)

PR #395, opened 2018-08-09
Add rake task to unpublish and redirect countries

rubenarakelyan:
This commit adds a new rake task which will unpublish and redirect a country’s travel advice and email signup pages to another country’s travel advice page.

Trello: https://trello.com/c/RkqvdT4z/377-remove-american-samoa-travel-advice

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -12,6 +12,14 @@ namespace :publishing_api do
     api_v2.publish(presenter.content_id)
   end
 
+  desc "unpublish a published edition and email signup content item for a country and redirect"
+  task :unpublish_published_edition_and_email_signup_content_item, %i(country_slug new_country_slug) => :environment do |_task, args|
+    country = Country.find_by_slug(args[:country_slug])
+    alternative_path = "/foreign-travel-advice/#{args[:new_country_slug]}"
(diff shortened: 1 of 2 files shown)


lib/tasks/publishing_api.rake
```diff
@@ -12,6 +12,14 @@ namespace :publishing_api do
     api_v2.publish(presenter.content_id)
   end
 
+  desc "unpublish a published edition and email signup content item for a country and redirect"
+  task :unpublish_published_edition_and_email_signup_content_item, [:country_slug, :new_country_slug] => :environment do |_task, args|
+    country = Country.find_by_slug(args[:country_slug])
+    alternative_path = "/foreign-travel-advice/#{args[:new_country_slug]}"
+    api_v2.unpublish(country.email_signup_content_id, type: "redirect", alternative_path: alternative_path)
```
thomasleese: Should we redirect this to the email sign up page of the new country?

rubenarakelyan: Yeah, I thought it may provide more insight if you were redirected to the main travel advice page, but I guess this provides more consistency. I'm not for or against either.

thomasleese: Yeah same... I can see both ways being confusing to different people...

PR #439, opened 2018-10-01
 Stop talking directly to email-alert-api

thomasleese:
We can rely on publishing-api and email-alert-service to send the emails for us.


cbaines:
I know from looking at the Specialist Publisher related subscriptions that the information in the Publishing API isn't sufficient to match content changes against subscriber lists. I haven't looked at the Travel Advice content though, will the way the content is formatted work with all the subscriptions (especially tags if they're used?)?

cbaines:
> @cbaines Yep, this is more complicated than I had initially thought, for that exact reason. I was hoping Travel Advice would be easier, but there's currently no way of subscribing to a single piece of content in Email Alert API which is what travel advice is. It works at the moment by the subscriber list having a link of `{ countries: [...]}` where the Content ID is just a reference to itself and Travel Advice Publisher manually includes this hash when it calls Email Alert API. This doesn't feel like something we should be adding in the Publishing API just to support Email Alerts.

Ok, thanks for looking in to it anyway!

PR #446, opened 2018-10-05
Add MIT licence file

andreagrandi:
Licence file was missing, so I added the one suggested on https://gds-way.cloudapps.digital/manuals/licensing.html#guidelines-for-repositories-containing-code

Cheers


thomasleese:
Added in #449.

PR #550, opened 2019-02-27
Archive american-samoa TravelAdviceEdition

AlanGabbianelli:
Trello: 
https://trello.com/c/7Ez4ivvM/811-archive-all-traveladviceeditions-for-american-samoa
Co-Authored-By: Elliot Crosby-McCullough <elliot.cm@gmail.com>

### DIFF
--- a/db/migrate/20190227101824_archive_american_samoa.rb
+++ b/db/migrate/20190227101824_archive_american_samoa.rb
@@ -1,0 +1,7 @@
+class ArchiveAmericanSamoa < Mongoid::Migration
+  def change
+    TravelAdviceEdition
+      .where(country_slug: "american-samoa", state: "published")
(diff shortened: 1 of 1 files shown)


db/migrate/20190227101824_archive_american_samoa.rb
```diff
@@ -0,0 +1,8 @@
+class ArchiveAmericanSamoa < Mongoid::Migration
+  def change
+    TravelAdviceEdition
+      .where(country_slug: "american-samoa", state: "published")
+      .first # there is only one
```
thomasleese: You could do `TravelAdviceEdition.find_by` here instead of `.where(...).first` which might be nicer.

Or I'd be tempted to not limit on the first one and just do `.where(...).update_all(state: "archived")`.

PR #744, opened 2019-11-21
Fix missing dependencies with rake tasks

theseanything:
Rake is unable to use the Rakefile when only production dependencies are installed. This is because the Rakefile uses RuboCop and scss_lint to define a rake task and is only available as a test dependency.


theseanything:
Closing as this is addressed by https://github.com/alphagov/travel-advice-publisher/pull/825/files

PR #750, opened 2019-11-29
Enable rails-specific rules for rubocop-govuk

benjamineskola:
trello: https://trello.com/c/Lijki8aj/1584-3-enable-rails-rubocop-rules-via-rubocop-govuk-on-all-our-rails-apps

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,18 @@
 inherit_gem:
   rubocop-govuk:
     - "config/default.yml"
+    - "config/rails.yml"
+
+# Mongoid doesn't support some of the methods that Rubocop prefers
+Rails/FindBy:
+  Enabled: false
+
+Rails/FindEach:
+  Enabled: false
+
+Rails/DynamicFindBy:
+  Whitelist:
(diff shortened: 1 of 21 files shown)


.rubocop.yml
```diff
@@ -13,3 +13,6 @@ Rails/FindEach:
 Rails/DynamicFindBy:
   Whitelist:
     - find_by_slug
+
+Rails/OutputSafety:
```
benjamineskola: I've disabled this here because the [recommended changes](https://github.com/rubocop-hq/rubocop-rails/blob/master/lib/rubocop/cop/rails/output_safety.rb) don't seem to work inside a FormBuilder. It might be something we want to revisit, though, as it's apparently a potential security risk.

benjamineskola:
These failures relate to a rule that has changed in rubocop-rails 2.4.0, which was released last week, so they'll be fixed by merging #749.

.rubocop.yml
AlanGabbianelli: I think it would be worth having this comment in the commit message itself.

PR #825, opened 2020-03-10
Rubocop fixes

JonathanHallam:
Platform health recently made some changes (https://github.com/alphagov/govuk-jenkinslib/pull/66) to how linting works that requires our repos to be re-linted. This fixes all of the issues that now show up.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,11 @@ inherit_mode:
   merge:
     - Exclude
 
+AllCops:
+  Exclude:
+    - db/migrate/201*.rb
+
+
 # Mongoid doesn't support some of the methods that Rubocop prefers
 Rails/FindBy:
   Enabled: false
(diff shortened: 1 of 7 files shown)


.rubocop.yml
```diff
@@ -7,6 +7,11 @@ inherit_mode:
   merge:
     - Exclude
 
+AllCops:
+  Exclude:
+    - db/migrate/201*.rb
```
issyl0: This doesn't seem to match with the later fixes you made to the `db/migrate/201*` files?

issyl0: Oh, sorry. It's just that you did this _after_ you autocorrected all the things, so there were changes to `db/migrate/201*` files anyway.

JonathanHallam: yeah basically, figured it was better just to fix the small stuff that it could and not invest extra time

PR #837, opened 2020-03-20
Clarify edition change form

laurentqro:
# What
In tandem with a push to ensure travel advice publishers have clear expectations about when emails will be sent - we can clarify the publishing interface to remind them of guidance / rules.

# Why
We have at least once case study of FCO sending 4 "minor" updates tagged as major to uses in a single day. This is a bad experience for users and leads to thousands of unnecessary emails.

# Solution

Whitehall's admin interface is very clear about when users will receive emails after existing content has been added. This PR brings travel advice publisher into line with Whitehall.

## Before

Screenshot 2020-03-30 at 14 50 37

## After

Screenshot 2020-03-30 at 14 49 39

# Checklist

* [x] Confirm copy

---

https://trello.com/c/XwFxyM8X/33-travel-advice-publisher-make-admin-interface-clearer-when-emails-will-be-sent

### DIFF
--- a/app/assets/javascripts/admin/editions/form.js
+++ b/app/assets/javascripts/admin/editions/form.js
@@ -1,0 +1,21 @@
+var adminEditionsForm = {
+  showChangeNotesIfMajorChange: function showChangeNotesIfMajorChange() {
+    var $form                      = $(".js-edition-form");
+    var $fieldset                  = $('.js-change-notes', this.$form);
+    var $radio_buttons             = $('input[type=radio]', $fieldset);
+    var $minor_change_radio_button = $('.js-edition-update-minor', $fieldset);
+    var $change_notes_section      = $('.js-change-notes-section', $fieldset);
+
+    $radio_buttons.change(showOrHideChangeNotes);
+    showOrHideChangeNotes();
+
+    function showOrHideChangeNotes() {
+      if ($minor_change_radio_button.prop('checked')){
+        $change_notes_section.slideUp(200);
+      }
+      else {
+        $change_notes_section.slideDown(200);
+      }
+    }
+  }
+}
--- a/app/assets/javascripts/travel_advice_utils.js
+++ b/app/assets/javascripts/travel_advice_utils.js
@@ -2,11 +2,4 @@ var TravelAdviceUtils = {
   convertToSlug: function(title) {
     return title.toLowerCase().replace(/[^\w ]+/g,'').replace(/ +/g,'-');
   },
-  setChangeDescriptionVisibility: function($elem) {
-    if ($elem.is(':checked')) {
-      $("#major_update_input").hide();
-    } else {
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/travel_advice_utils.js
```diff
@@ -2,11 +2,23 @@ var TravelAdviceUtils = {
...
+
+    if ($minorUpdate.is(':checked')) {
+      $majorUpdateText.hide();
     } else {
-      $("#major_update_input").show();
+      $majorUpdateText.show();
     }
-  }
+
+    $minorUpdate.change(function() {
```
vanitabarrett: I think we could tidy up the change listeners here, Whitehall actually has a pretty good pattern for this(!).
https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/app/assets/javascripts/admin/views/editions/_form.js#L27

If we follow that pattern, we can also use it to set the initial state and therefore remove the 'is checked' check above.

app/views/admin/editions/edit.html.erb
```diff
@@ -20,22 +20,29 @@
...
-                <p>If the change is major, leave a description below. An email will be sent to subscribers with your description of the change.</p>
-              <% end %>
-              <%= f.input :change_description, :as => :text, :label => "Change description (plain text)", :required => true, :input_html => {
-                :class => "input-md-8", :rows => 4, :disabled => ! @edition.draft? } %>
-            </div>
-          <% end %>
+              <div class="radio">
+                <%= label_tag do %>
+                  <%= f.radio_button(:minor_update, false, class: "js-edition-update-major") %>
+                  <strong>A significant change, for example a new travel ban</strong> (sends an email to all subscribers and changes the "latest update" description on the summary page)
```
karlchillmaid:
```suggestion
                  <strong>A significant change, for example a new travel ban</strong> (sends an email to all subscribers and adds a change note to the summary page)
```

app/views/admin/editions/edit.html.erb
```diff
@@ -20,22 +20,29 @@
...
+                <% end %>
+              </div>
+
+              <div class="form-group js-change-notes-section">
+                <%= f.label :change_description, "Latest update" %>
```
karlchillmaid:
```suggestion
                <%= f.label :change_description, "Change note" %>
```

vanitabarrett:
@laurentqro Code change looks good, there's a couple of tests that need updating

app/assets/javascripts/admin/editions/form.js
```diff
@@ -0,0 +1,21 @@
+var adminEditionsForm = {
```
vanitabarrett: I'm not sure this javascript file needs to live under admin/editions? It could probably do with a more descriptive filename too.

PR #908, opened 2020-06-22
Replace minor_update boolean attribute with update_type of type string

laurentqro:
### What

This PR is to replace `TravelAdviceEdition`'s `minor_update` boolean attribute with an `update_type` of type `string`.

### Why

While I understand why a boolean field can make sense here, it also means we are unable to make full use of some Rails freebies, like validation error messages. `Minor update can't be blank` is indeed somewhat confusing.

This is preparation work for https://github.com/alphagov/travel-advice-publisher/pull/900.

### DIFF
--- a/app/models/travel_advice_edition.rb
+++ b/app/models/travel_advice_edition.rb
@@ -17,6 +17,7 @@ class TravelAdviceEdition
   field :summary,              type: String,    default: ""
   field :change_description,   type: String
   field :minor_update,         type: Boolean,   default: false
+  field :update_type,          type: String,    default: "major"
   field :synonyms,             type: Array,     default: []
   # This is the publicly presented publish time. For minor updates, this will be the publish time of the previous version
   field :published_at,         type: Time
@@ -57,7 +58,7 @@ class TravelAdviceEdition
 
   state_machine initial: :draft do
     before_transition draft: :published do |edition, _|
-      if edition.minor_update
+      if edition.is_minor_update?
         previous = edition.previous_version
         edition.published_at = previous.published_at
         edition.reviewed_at = previous.reviewed_at
@@ -79,11 +80,15 @@ class TravelAdviceEdition
 
     state :published do
       validate :cannot_edit_published
-      validates :change_description, presence: { unless: :minor_update, message: "can't be blank on publish" }
+      validates :change_description, presence: { unless: :is_minor_update?, message: "can't be blank on publish" }
     end
     state :archived do
       validate :cannot_edit_archived
     end
+  end
+
+  def is_minor_update?
+    update_type == "minor" || minor_update
   end
 
   def build_clone(target_class = nil)
@@ -107,7 +112,7 @@ class TravelAdviceEdition
   end
 
   def publish_as(user)
-    comment = minor_update ? "Minor update" : Govspeak::Document.new(change_description).to_text
+    comment = is_minor_update? ? "Minor update" : Govspeak::Document.new(change_description).to_text
     build_action_as(user, Action::PUBLISH, comment) && publish
   end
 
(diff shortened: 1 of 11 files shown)


app/models/travel_advice_edition.rb
```diff
@@ -79,13 +80,17 @@ class << self; attr_accessor :fields_to_clone end
...
-      validates :change_description, presence: { unless: :minor_update, message: "can't be blank on publish" }
+      validates :change_description, presence: { unless: :is_minor_update?, message: "can't be blank on publish" }
     end
     state :archived do
       validate :cannot_edit_archived
     end
   end
 
+  def is_minor_update?
+    minor_update || update_type == "minor"
```
benthorner, line 91:
```suggestion
    minor_update || minor?
```

Minor: you can use [an enum](https://github.com/alphagov/content-publisher/blob/master/app/models/status.rb#L16) here; this could wait until we've got rid of the original column, so it's not too confusing.

app/views/admin/editions/_change_notes.html.erb
```diff
@@ -3,15 +3,15 @@
     <% if @edition.update? %>
       <div class="radio">
         <%= label_tag do %>
-          <%= f.radio_button(:minor_update, true, class: "js-edition-update-minor") %>
+          <%= f.radio_button(:update_type, "minor", class: "js-edition-update-minor") %>
```
benthorner, line 6: We should make sure this populates with the value of the old column, if present. The simplest way might be to use the old behaviour, but with `is_minor_update?`.

laurentqro: So we want to read the value of the old column if present, but we also want to write to the new one. Am I understanding this correctly?

app/models/travel_advice_edition.rb
laurentqro: Am I able to use Rails' enum even though Mongoid doesn't have an `enum` type?

benthorner: Oops, you're right - I am thinking of Postgres integer/enum.

app/views/admin/editions/_change_notes.html.erb
```diff
@@ -3,15 +3,15 @@
     <% if @edition.update? %>
       <div class="radio">
         <%= label_tag do %>
-          <%= f.radio_button(:minor_update, true, class: "js-edition-update-minor") %>
+          <%= f.radio_button(:update_type, "minor", checked: @edition.is_minor_update?, class: "js-edition-update-minor") %>
```
benthorner, line 6: Sorry: I think there's still a problem with this:  a user won't be able to change an existing minor update to major, because we'll be changing the new column, which [has lower precedence than the unchanged, boolean original](https://github.com/alphagov/travel-advice-publisher/pull/908/files#diff-38443ace3e674824c89321de36d200b5R91). 

We could address this issue by reversing [the precedence](https://github.com/alphagov/travel-advice-publisher/pull/908/files#diff-38443ace3e674824c89321de36d200b5R91) of the columns.

spec/features/edition_edit_spec.rb
```diff
@@ -267,7 +267,7 @@ def reorder_parts(index, new_index)
     click_navbar_button "Save"
 
     @edition.reload
-    expect(@edition.minor_update).to eq(false)
+    expect(@edition.is_minor_update?).to eq(false)
```
benthorner, line 270:
```suggestion
    expect(@edition.update_type).to eq("minor")
```

Minor: the rest of the tests check the value directly.

spec/models/travel_advice_edition_spec.rb
```diff
@@ -260,7 +260,7 @@
     end
 
     it "is not minor_update" do
-      expect(TravelAdviceEdition.new.minor_update).to be false
+      expect(TravelAdviceEdition.new.is_minor_update?).to be false
```
benthorner, line 263:
```suggestion
      expect(TravelAdviceEdition.new.update_type).to_not eq("minor")
```

(also this one)

laurentqro:
thanks for the review @benthorner , I made the tweaks you recommended

PR #911, opened 2020-06-23
Migrate travel advice edition records to new update_type field

laurentqro:
To be merged and deployed shortly after https://github.com/alphagov/travel-advice-publisher/pull/908.

https://trello.com/c/LWl84KOC/336-dont-pre-check-a-significant-change-for-example-a-new-travel-restriction-when-publishers-create-new-editions

### DIFF
--- a/db/migrate/20200623144214_change_minor_update_boolean_to_update_type_string.rb
+++ b/db/migrate/20200623144214_change_minor_update_boolean_to_update_type_string.rb
@@ -1,0 +1,11 @@
+class ChangeMinorUpdateBooleanToUpdateTypeString < Mongoid::Migration
+  def self.up
+    TravelAdviceEdition.where(minor_update: true, update_type: nil).update_all(update_type: "minor")
+    TravelAdviceEdition.where(minor_update: false, update_type: nil).update_all(update_type: "major")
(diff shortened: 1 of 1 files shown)


db/migrate/20200623144214_change_minor_update_boolean_to_update_type_string.rb
```diff
@@ -0,0 +1,11 @@
+class ChangeMinorUpdateBooleanToUpdateTypeString < Mongoid::Migration
+  def self.up
+    TravelAdviceEdition.where(minor_update: true, update_type: nil).update_all(update_type: "minor")
+    TravelAdviceEdition.where(minor_update: false, update_type: nil).update_all(update_type: "major")
+  end
+
+  def self.down
+    TravelAdviceEdition.where(update_type: "minor").update_all(minor_update: true)
```
benthorner, line 8:
```suggestion
    TravelAdviceEdition.where(update_type: "minor").update_all(minor_update: true, update_type: nil)
```

Pretty minor, but we wouldn't be able to run the forward migration correctly again if we had to rollback, since we would no longer have `update_type: nil` as a flag.

PR #937, opened 2020-08-03
Remove usage of `mongoid_rails_migrations` gem & all migrations

issyl0:
- To [pay down some tech debt](https://trello.com/c/CUj0jrkN/184-were-using-an-old-forked-gem-for-mongodb-migrations-that-isnt-necessary), we're removing this gem and all data migrations.
- While this repo does not use our forked branch of the gem (like other apps did), we should still tidy its use up for consistency with all of our MongoDB applications.
- MongoDB is schemaless, and one would not expect to see migrations in a MongoDB project.
- All of these migrations were data migrations from 2013 to 2020, and most likely (tho untested) won't apply from scratch any more (this is why we rely on replicated data).
- Our preferred approach for data changes these days is to write short-lived Rake tasks.

https://trello.com/c/CUj0jrkN/184-were-using-an-old-forked-gem-for-mongodb-migrations-that-isnt-necessary


thomasleese:
We'll need to update the docs here to come up with a new way of renaming a country: https://docs.publishing.service.gov.uk/manual/rename-a-country.html#1-update-travel-advice-publisher

It sounds reasonable to me that we could have a Rake task `country:rename[old-slug,new-slug]` that does what the migrations used to do.

issyl0:
@thomasleese Ahhh, that explains things. I added a Rake task.

lib/tasks/rename_country.rake
```diff
@@ -0,0 +1,11 @@
+namespace :country do
+  desc "Rename a country"
+  task :rename, %i[old_country_slug new_country_slug] => :environment do |_task, args|
+    if args.old_country_slug.blank? || args.new_country_slug.blank?
```
thomasleese, line 4:
```suggestion
    if args[:old_country_slug].blank? || args[:new_country_slug].blank?
```

I may be wrong, but I _think_ this is how to reference arguments in Rake tasks.

issyl0: It appears to work either way. 🤷🏻

PR #1299, opened 2022-04-05
Migrate list and show page to GOVUK Design systems

DilwoarH:
## What
Migrate the List and show page to the GOVUK design system

## Why
Migrating to GOVUK Design System will make it easier for users to use, consistent with the rest of GOVUK, and more accessible

## Before

Screenshot 2022-04-06 at 3 15 53 am
Screenshot 2022-04-06 at 3 15 40 am

## After

Screenshot 2022-04-06 at 3 15 10 am
Screenshot 2022-04-06 at 3 15 24 am

https://trello.com/c/sFromAP6/352-travel-advice-list-index-page

### DIFF
--- a/app/controllers/admin/countries_controller.rb
+++ b/app/controllers/admin/countries_controller.rb
@@ -1,15 +1,26 @@
 class Admin::CountriesController < ApplicationController
-  layout "legacy"
+  layout :get_layout
   before_action :skip_slimmer
   before_action :load_country, only: [:show]
 
   def index
     @countries = Country.all
+    render "admin/countries/index_legacy" if is_legacy_layout?
   end
 
   def show; end
 
 private
+
+  def is_legacy_layout?
+    !preview_design_system_user?
+  end
+
+  def get_layout
+    return "legacy" if is_legacy_layout? || %(show).include?(action_name)
+
+    "design_system"
+  end
 
   def load_country
     @country = Country.find_by_slug(params[:id]) || error_404
--- a/app/controllers/admin/countries_controller.rb
+++ b/app/controllers/admin/countries_controller.rb
@@ -8,7 +8,9 @@ class Admin::CountriesController < ApplicationController
     render "admin/countries/index_legacy" if is_legacy_layout?
   end
 
-  def show; end
+  def show
+    render "admin/countries/show_legacy" if is_legacy_layout?
+  end
 
 private
 
@@ -17,7 +19,7 @@ private
   end
 
   def get_layout
-    return "legacy" if is_legacy_layout? || %(show).include?(action_name)
+    return "legacy" if is_legacy_layout?
 
     "design_system"
   end
--- a/app/views/admin/countries/index.html.erb
+++ b/app/views/admin/countries/index.html.erb
@@ -1,34 +1,18 @@
-<h1 class="page-title">
-  Countries
-</h1>
+<% content_for :title, "Countries" %>
 
-<table class="table table-striped table-bordered" data-module="filterable-table">
-  <thead>
-    <tr class="table-header">
-      <th>Country</th>
-      <th>Status</th>
-    </tr>
-    <!--[if gte IE 8]><!-->
-    <tr class="if-no-js-hide table-header-secondary">
-      <td colspan="2">
-        <form>
-          <label for="table-filter" class="rm">Filter countries</label>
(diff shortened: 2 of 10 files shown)


app/views/admin/countries/_show-legacy.html.erb
```diff
@@ -0,0 +1,40 @@
+<ul class="breadcrumb">
```
davidgisbey, line 1: Normally partials use underscores rather than dashes in the file name

app/views/admin/countries/index.html.erb
```diff
@@ -1,34 +1,23 @@
-<h1 class="page-title">
-  Countries
-</h1>
+<%= render :partial => 'index-legacy' if @legacy_layout %>
```
davidgisbey, line 1: Not a massive deal, but I think it's a bit cleaner to handle this in the controller. I'll leave a comment there

app/controllers/admin/countries_controller.rb
```diff
@@ -1,16 +1,29 @@
...
   before_action :load_country, only: [:show]
 
   def index
     @countries = Country.all
+    @legacy_layout = is_legacy_layout?
   end
 
-  def show; end
+  def show
+    @legacy_layout = is_legacy_layout?
```
davidgisbey, line 12: render `show_legacy` if is_legacy_layout?

Then change move the partial into a new view

app/controllers/admin/countries_controller.rb
```diff
@@ -1,16 +1,29 @@
 class Admin::CountriesController < ApplicationController
-  layout "legacy"
+  layout :get_layout
   before_action :skip_slimmer
   before_action :load_country, only: [:show]
 
   def index
     @countries = Country.all
+    @legacy_layout = is_legacy_layout?
```
davidgisbey, line 8: render `show_index` if is_legacy_layout?

Then change move the partial into a new view. That way the views are decoupled.

Wdyt?

app/controllers/admin/countries_controller.rb
```diff
@@ -1,16 +1,29 @@
...
 
-  def show; end
+  def show
+    @legacy_layout = is_legacy_layout?
+  end
 
 private
 
+  def is_legacy_layout?
+    ! preview_design_system_user?
```
davidgisbey, line 18: `!preview_design_system_user?`

in ruby

app/controllers/admin/countries_controller.rb
DilwoarH: good idea

app/views/admin/countries/_show-legacy.html.erb
DilwoarH: done

app/views/admin/countries/index.html.erb
DilwoarH: good idea - done

app/controllers/admin/countries_controller.rb
DilwoarH: done

app/controllers/admin/countries_controller.rb
DilwoarH: done

davidgisbey (approved):
lgtm

davidgisbey:
Yea all good. I've gone for a slightly different approach https://github.com/alphagov/travel-advice-publisher/pull/1302/commits/f84f92c0f43d0812e55aa6e530ec19cde9845f76

where i added some helper methods and split the specs out into different files (mainly because the specs are already massive so it'll be easier to delete them). I don't think it's overly necessary with yours as there are so few tests but up to you.

DilwoarH:
> Yea all good. I've gone for a slightly different approach [f84f92c](https://github.com/alphagov/travel-advice-publisher/commit/f84f92c0f43d0812e55aa6e530ec19cde9845f76)
> 
> where i added some helper methods and split the specs out into different files (mainly because the specs are already massive so it'll be easier to delete them). I don't think it's overly necessary with yours as there are so few tests but up to you.

Which approach do you think we should follow?

I followed the approach used in content publisher: https://github.com/alphagov/content-publisher/blob/c8d93843e99318d4180ce93ed4e07abb03395121/spec/requests/editions_spec.rb#L29-L31

davidgisbey:
Generally if you're reusing a bit of code repeatedly throughout multiple spec files you add a helper method. If it's just once or twice then adding it like Content Publisher is fine. Having said that it's somewhat nit picky and not a blocking change.

PR #1300, opened 2022-04-06
Add new design system user permission helper

DilwoarH:
## What

Add new permission to allow users to see the design system.

## Why

This will allow users to test design system changes before we deploy them out to everyone (like a feature flag).

## How to test

To test this you will need to reseed your database, the best way to do this is by the following command (this will reset your db):

```sh
govuk-docker run travel-advice-publisher-app bundle exec rails db:reset
```

https://trello.com/c/jUeXetdL/359-add-new-design-system-user-permission

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,6 +11,12 @@ class ApplicationController < ActionController::Base
 
 private
 
+  helper_method :preview_design_system_user?
+
+  def preview_design_system_user?
+    current_user.has_permission? "Preview Design System"
+  end
+
   def error(status_code)
     render status: status_code, plain: "#{status_code} error"
   end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -3,7 +3,7 @@ if User.where(name: "Test user").blank?
 
   User.create!(
     name: "Test user",
-    permissions: %w[signin gds_editor],
+    permissions: ["signin", "GDS Editor", "Preview Design System"],
     organisation_content_id: gds_organisation_id,
   )
 end


kevindew (approved):
This looks good.

I'm trying to think if there are scenarios where current_user might be nil (I know in the past they can occur on error pages for instance) but I think you'll be ok. A way to resolve that would be using safe navigation operator e.g: `current_user&.has_permission?("Preview Design System")`

davidgisbey (approved):
lgtm

DilwoarH:
> This looks good.
> 
> I'm trying to think if there are scenarios where current_user might be nil (I know in the past they can occur on error pages for instance) but I think you'll be ok. A way to resolve that would be using safe navigation operator e.g: `current_user&.has_permission?("Preview Design System")`

I'll keep this in mind if we start seeing issues. At the moment, I don't see why a user would be on the app without permission.

kevindew:
> > This looks good.
> > I'm trying to think if there are scenarios where current_user might be nil (I know in the past they can occur on error pages for instance) but I think you'll be ok. A way to resolve that would be using safe navigation operator e.g: `current_user&.has_permission?("Preview Design System")`
> 
> I'll keep this in mind if we start seeing issues. At the moment, I don't see why a user would be on the app without permission.

Yeah, like I say it tends to happen for error pages in generally and I imagine we've only got the Rails default ones now. But I did also notice the method below checks if `current_user` exists. I mostly mention it because this expectation has caught us out in the past: https://github.com/alphagov/content-publisher/commit/0922b7ddad959f2195765df14a9e456093baa43b but I am quite confident you won't have an issue.

PR #1304, opened 2022-04-08
Move adding parts to a separate flow - Design System port

davidgisbey:
## Description

The current implementation of adding parts uses Javascript for adding, removing and reordering parts and is on the same page as editing the rest of the edition.

It's also completely inaccessible at it doesn't work without JS. You error out when you try and add a part.

In order to get round using heavy JS, we can follow the more modern publishing applications example and move adding, editing, removing and reordering parts into its own flow. There's prior art for the exact component we need in the step by step interface in publisher https://collections-publisher.integration.publishing.service.gov.uk/step-by-step-pages/87

## Changes made 

This PR does the following

1. Adds endpoints for adding a new part
2. Updates the parts partial on the edit page to use a summary list component rather than an accordion
3. Adds an error summary banner partial and an error helper method for linking errors to inputs (the same implementation as was used in content publisher
4. Adds feature and controller specs for adding a new part 

### Screenshots 

New page 

image

Edit edition page 

|Before|After|
|------------|------------|
|image|image|

### Out of scope 
- Editing a part
- Deleting a part 
- Reordering a part

These will come in follow up PRs

### Trello card 

https://trello.com/c/E48t5WGw/365-travel-advice-move-adding-and-editing-govspeak-parts-to-their-own-flow


brucebolt (approved):

app/views/admin/editions/_parts.html.erb
```diff
@@ -1,61 +1,23 @@
...
-        <div class="govuk-accordion__section-header">
-          <h2 class="govuk-accordion__section-heading">
-            <span class="govuk-accordion__section-button" id=<%= part.title.dasherize || "untitled" %>>
-              <%= part.title.present? ? part.title : 'Untitled part' %>
-            </span>
-          </h2>
-        </div>
-        <div id="<%= part.slug || 'untitled-part' %>" class="govuk-accordion__section-content" aria-labelledby="<%= part.title.dasherize || "untitled" %>">
-          <% #TODO: all of these are required fields and need to be wired up to produce errors %>
+<div id="parts" class="govuk-!-margin-bottom-3">
```
DilwoarH, line 1: do we need an ID here?

app/views/admin/editions/_parts.html.erb
```diff
@@ -1,61 +1,23 @@
...
-  <%= render "govuk_publishing_components/components/button", {
-    text: "Add new part"
-  } %>
-<% end %>
+  <% if @edition.draft? %>
+    <%= render "govuk_publishing_components/components/button", {
+      text: "Add part",
+      href: new_admin_country_edition_part_path(@country.slug, @edition)
+    } %>
+  <% end %>
```
DilwoarH, line 22: We should probably make this a link rather than a button

spec/factories.rb
```diff
@@ -20,6 +20,10 @@
     trait :with_design_system_permission do
       permissions { ["signin", "Preview Design System"] }
     end
+
+    trait :with_design_system_permission do
+      permissions { ["signin", "Preview Design System"] }
+    end
```
DilwoarH, line 26:
```suggestion
```

spec/factories.rb
```diff
@@ -20,6 +20,10 @@
     trait :with_design_system_permission do
       permissions { ["signin", "Preview Design System"] }
     end
+
+    trait :with_design_system_permission do
```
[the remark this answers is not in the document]
DilwoarH: removed

app/views/admin/editions/_parts.html.erb
DilwoarH: checked we need it for tests

app/views/admin/editions/_parts.html.erb
DilwoarH: made it into secondary button

PR #1308, opened 2022-04-12
Adds CSRF token to create new edition form

DilwoarH:
During the upgrade, the rails post button was converted to a form. This meant that the csrf field was missed out as it was previously added automatically.

This change adds it back in.

### DIFF
--- a/app/views/admin/countries/show.html.erb
+++ b/app/views/admin/countries/show.html.erb
@@ -16,12 +16,12 @@
 
 <% content_for :title_side do %>
   <% unless @country.has_draft_edition? %>
-    <form action="<%= admin_country_editions_path(@country.slug) %>" method="POST">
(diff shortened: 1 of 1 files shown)


app/views/admin/countries/show.html.erb
```diff
@@ -16,12 +16,12 @@
 
 <% content_for :title_side do %>
   <% unless @country.has_draft_edition? %>
-    <form action="<%= admin_country_editions_path(@country.slug) %>" method="POST">
+    <% form_with url: admin_country_editions_path(@country.slug), method: :post do |f| %>
```
davidgisbey, line 19: sorry i missed the = 

```
 <%= form_with url: admin_country_editions_path(@country.slug), method: :post do |f| %>

```

DilwoarH: ops

davidgisbey (approved):

PR #1318, opened 2022-04-13
Move historical editions to design system

DilwoarH:
## What
Migrate the Historical editions to the GOVUK design system

## Why
Migrating to DS will make it easier for users to use, consistent with the rest of GOVUK, and more accessible

## Other notes
Ignore the print page

https://trello.com/c/hCLCz4cl/356-travel-advice-historical-editions-page


DilwoarH:
This isn't needed as historical editions uses it's own layout (not bootstrap but print style)

PR #1319, opened 2022-04-13
Add tracking back in for Travel advice publisher

DilwoarH:
## What
Add Google analytics tracking back into the application layout.

## Why
During the migration we lost the automatic analytics tracking, this needs to be added back in.

## Additional Information
Add this to the layout so that everything is tracked automatically.

https://trello.com/c/YlaNbIF6/373-add-tracking-back-in-for-travel-advice-publisher

### DIFF
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -1,0 +1,12 @@
+(function (GOVUK) {
+  'use strict'
+
+  GOVUK.Analytics.load()
+  GOVUK.analyticsVars = { primaryLinkedDomains: [document.domain] }
+  GOVUK.analytics = new GOVUK.Analytics({
+    universalId: 'UA-26179049-6', // GOVUK Apps GA ID
+    cookieDomain: document.domain
+  })
+
+  GOVUK.analytics.trackPageview()
+})(window.GOVUK)
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,2 +1,4 @@
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/all_components
+//= require govuk_publishing_components/analytics
+//= require analytics


app/assets/javascripts/analytics.js
```diff
@@ -0,0 +1,20 @@
+(function (GOVUK) {
+  'use strict'
+
+  // Load Google Analytics libraries
+  GOVUK.Analytics.load()
+
+  // Initialise analyticsVars
+  GOVUK.analyticsVars = window.GOVUK.analyticsVars || {}
+  GOVUK.analyticsVars.primaryLinkedDomains = [document.domain]
```
kevindew, line 9: Is there any reason why we'd expect this to be set already in this app? Not sure why we're not doing: `GOVUK.analyticsVars = { primaryLinkedDomains: [document.domain] }` since we then overwrite GOVUK.analytics on the next command

app/assets/javascripts/analytics.js
```diff
@@ -0,0 +1,20 @@
+(function (GOVUK) {
+  'use strict'
+
+  // Load Google Analytics libraries
```
kevindew, line 4: This doesn't seem a useful comment, just seems to rephrase what the code says

app/assets/javascripts/analytics.js
```diff
@@ -0,0 +1,20 @@
...
+  GOVUK.analyticsVars.primaryLinkedDomains = [document.domain]
+
+  // Configure profiles and make interface public
+  // for custom dimensions, virtual pageviews and events
+  GOVUK.analytics = new GOVUK.Analytics({
+    universalId: 'UA-26179049-6', // GOVUK Apps GA ID
+    cookieDomain: document.domain
+  })
+
+  // Track initial pageview
```
kevindew, line 18: This comment also seems to just say what the code says

app/assets/javascripts/analytics.js
```diff
@@ -0,0 +1,20 @@
...
+  GOVUK.Analytics.load()
+
+  // Initialise analyticsVars
+  GOVUK.analyticsVars = window.GOVUK.analyticsVars || {}
+  GOVUK.analyticsVars.primaryLinkedDomains = [document.domain]
+
+  // Configure profiles and make interface public
+  // for custom dimensions, virtual pageviews and events
+  GOVUK.analytics = new GOVUK.Analytics({
+    universalId: 'UA-26179049-6', // GOVUK Apps GA ID
```
kevindew, line 14: Do we just hardcode this in and run it everywhere, even local machines? Though can see it in govuk_admin_template https://github.com/alphagov/govuk_admin_template/blob/d2a0be75c1931fca2d5e292f7c8017047768f763/app/views/layouts/govuk_admin_template.html.erb#L119 so guess we've been happy with that.

We used to set this id in Content Publisher with an env var: https://github.com/alphagov/content-publisher/blob/3f14860c12aae112f0da2afb5314740d253177d2/app/views/layouts/application.html.erb#L16-L22

kevindew (commented):
This looks good, just a few minor comments.

There seems to be some clear things that could be improved with this (not having all analytics use same id in every env, moving some of this initialisation into the GOVUK.Analytics library). But doesn't seem worth pursing those improvements this late in universal analytics life.

app/assets/javascripts/analytics.js
DilwoarH: I followed the [documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics/analytics.md#create-an-analytics-tracker) however, looking at this again and the above examples, it might be better to add it to the [admin layout component ](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb)

kevindew: It'd be dangerous to do that as I know of apps, eg Content Publisher, that use it but have Google Tag Manager integration

DilwoarH: The only question I have is that we'd want these analytics events to run locally so that we can add new tracking more easily and not just run it in production like the above example. Is there an environment variable that I can use that would normally hold the GA ID?

DilwoarH: > It'd be dangerous to do that as I know of apps, eg Content Publisher, that use it but have Google Tag Manager integration

In that case, I think the current solution is good. I'll tidy up the comments.

DilwoarH: I do think we should standardise our tracking across our admin apps and not have special cases like Content Publisher

kevindew: We'd have to decide on one, but as my earlier comment, it doesn't seem worth it at this point in Universal Analytics life.

In general just seems a bit of a mess to be sending analytics data in all local dev (presumably every execution of the test suite would also send over data) so would just add a ton of stuff all the time for just the very occasional benefit of testing a new tracking locally.

Where analytics runs seems like something the people sorting out GA4 should have an opinion on, probably worth chatting to Harriet when she's back.

DilwoarH: I'll give Harriet a shout once she is back.

kevindew (approved):
Thanks

PR #1337, opened 2022-05-09
Disable Fields when not editable

DilwoarH:
##What
When editions are not editable, disable the field

##Why
This is needed to stop users from editing the form as it has no effect (which may be misleading to users)

https://trello.com/c/rGUj0EfJ/420-disable-fields-when-not-editable

## Before

travel-advice-publisher dev gov uk_admin_editions_6278f4e3a195800001afd836_edit(iPad Pro) (1)

## After

travel-advice-publisher dev gov uk_admin_editions_6278f4e3a195800001afd836_edit(iPad Pro)


davidgisbey:
Looks good. Might be worth adding a test to check the fields aren't present for non-draft additions, but looks great 👍

DilwoarH:
> Looks good. Might be worth adding a test to check the fields aren't present for non-draft additions, but looks great 👍

done

PR #1338, opened 2022-05-10
Alert statuses now returning a list of items

DilwoarH:
Previously it would return an array of items but this isn't very nice or accessible.

This change fixes this.

Screenshot 2022-05-10 at 10 38 47 am


kevindew:
Will this look ok when there are no alert statuses checked?

DilwoarH:
> Will this look ok when there are no alert statuses checked?

Screenshot 2022-05-10 at 12 46 09 pm

It was empty before but I've added in some placeholder text

kevindew:
Yup that looks good to me

kevindew (approved):

PR #1479, opened 2022-11-16
test out latest jenkins lib

BeckaL:
This demonstrates that we can use the [proposed changes to jenkinslib](https://github.com/alphagov/govuk-jenkinslib/pull/126) without an issue.

This branch maintains the behaviour of cloning schemas from govuk-content-schemas, rather than publishing api, so demonstrates that the change to jenkinslib is safe to merge and will not cause repos that are currently configured to get schemas from govuk-content-schemas to fail, while we do the work to switch over.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,6 @@
 #!/usr/bin/env groovy
 
-library("govuk")
+library("govuk@add-publishing-api-clone")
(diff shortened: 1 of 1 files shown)


ollietreend:
No longer needed - tidying up ✨🧹

PR #1599, opened 2023-05-16
Fix for edition diff comparison not working when both summaries are nil

minhngocd:
Previously, all editions have summaries - even in cases where nothing is entered, it will be empty string. 

Since summary has always been required and defaulted to empty string, edition comparison always worked. 
However, since we have migrated summaries into parts, summary has become nil for all new published editions. 

When we compare older versions with summaries it was fine, however comparing newest versions where both summaries are nil presented an error. This fix makes sure that diffs work for nil fields, and also the comparison section for Summary will not show if neither compared versions have a summary.

How to recreate problem before fix: 
1. Run migration task for <country A> to move summary into parts
2. Create a new edition and publish for <country A>
3. Compare the diff between these two versions, both summaries will be nil

After fix: 
1. Diff for edition where Summary section is moved into Parts will show 2 empty summary sections if the Travel Advice had no summary before nor after the migration. If there were text, then it would show the diff of removal and addition.
Screenshot 2023-05-16 at 17 10 11

2. For subsequent versions it will no longer show Summary as a section in the diff if neither versions have existing summaries
Screenshot 2023-05-16 at 17 12 06

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -37,6 +37,6 @@ module ApplicationHelper
   end
 
   def diff_html(version1, version2)
-    Diffy::Diff.new(version1, version2, allow_empty_diff: false).to_s(:html).html_safe
+    Diffy::Diff.new(version1.to_s, version2.to_s, allow_empty_diff: false).to_s(:html).html_safe
   end
 end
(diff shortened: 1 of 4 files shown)


ollietreend (approved):
Nice work @minhngocd! 🚀 This looks good to me. And nice touch removing the 'summary' field when it's no longer relevant.

PR #1804, opened 2024-02-05
Remove unused publication checks

beccapearce:
This code is not used anywhere, let's get rid of it.

The model Publish request in production does not have any data filled in beyond the default values and it seems this is a check that is no longer used.

Trello card: https://trello.com/c/2YjMbfcg/2332-travel-advice-publisher-improvements-legacy-code-clean-up


beccapearce:
Should we be dropping the table for Publishing requests too? I'd be inclined to, but maybe we could leave it for a couple of weeks just in case there is any issue

ryanb-gds (approved):
This is really interesting. It looks like this was an attempt at acknowledging the fact that we don't really know what has happened after a publishing app makes a call to publishing API. I assume the rake task would run on a regular schedule to try to rectify things if the publishing API hadn't succeeded in publishing. I take it we've checked there isn't a scheduled task in the helm charts running that rake task?

We may have to solve this problem ourselves if we move the scheduling logic into Publishing API, but I feel a webhook from Publishing API back to the publishing app might be a better approach.

beccapearce:
So I think that when this was written the publishing infrastructure was a bit different, travel advice had its own front end that we weren't sure it was making it into which is part of the reason I don't think it's used anymore.

And yes I had a look and couldn't find the job running anywhere, or see it mentioned as a fix for anything.

baisa (approved):

baisa (commented):
Nice clean up @beccapearce, it will super helpful to merge it before we start scheduling implementation ❤️

PR #1831, opened 2024-03-08
Add scheduling

lauraghiorghisor-tw:
[Trello card](https://trello.com/c/CoAmIXCO/2333-travel-advice-publisher-improvements-scheduling-implementation-create-flow)


app/controllers/admin/editions_controller.rb
```diff
@@ -125,6 +127,17 @@ def save_and_publish
     end
   end
 
+  def save_and_schedule
+    redirect_to admin_countries_path and return unless can_schedule_edition?
+
+    if @edition.update(permitted_edition_attributes) && @edition.has_valid_change_description_for_scheduling?
```
lauraghiorghisor-tw, line 133: This `has_valid_change_description_for_scheduling?` -> we have validation on the state in the model anyway, which would produce an error on the scheduling form submit, but to keep the same pattern we have for "Save & Publish", we thought it would be nice to error before we redirect to `/schedulings/new`

app/workers/publish_scheduled_edition_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+      notifier.send_alert(edition)
+      notifier.enqueue
+
+      user = User.find(user_id)
+      edition.publish_as(user)
+    else
+      Sidekiq.logger.info("Scheduled published time should be in the past.")
+    end
+  rescue StandardError => e
+    raise WorkerError.new(self, e, "Edition must be in a scheduled state")
```
lauraghiorghisor-tw, line 21: Now, this error is misleading.

spec/models/travel_advice_edition_spec.rb
```diff
@@ -365,6 +400,88 @@
...
+      end
+    end
+
+    it "sets the change_description to the previous version's change_description for a minor update" do
+      scheduled.update_type = "minor"
+      scheduled.publish!
+      expect(scheduled.change_description).to eq(published.change_description)
+    end
+
+    context "#schedule_for_publication" do
```
lauraghiorghisor-tw, line 447: Used to have these in the controller tests. Moved them here, kept the controller tests slimmer.Used to have these in the controller tests. Moved them here, kept the controller tests slimmer.

app/models/travel_advice_edition.rb
```diff
@@ -107,15 +119,25 @@ def build_clone(target_class = nil)
...
   def publish_as(user)
     comment = is_minor_update? ? "Minor update" : Govspeak::Document.new(change_description).to_text
     build_action_as(user, Action::PUBLISH, comment) && publish
   end
 
+  def schedule_for_publication(user)
+    return false unless build_action_as(user, Action::SCHEDULE_FOR_PUBLICATION, nil, scheduled_publication_time:) && schedule
+
+    notifier.put_content(self)
+    notifier.patch_links(self)
```
lauraghiorghisor-tw, line 135: Not sure if we require the patch links too.

spec/requests/request_tracing_spec.rb
```diff
@@ -8,6 +8,8 @@
   let(:edition) { create(:travel_advice_edition, country_slug: "aruba") }
 
   before do
+    Sidekiq::Worker.clear_all
```
lauraghiorghisor-tw, line 11: Hoping this fixes it. 
Think the drain_all in the test was getting the worker to run and it couldn't find the edition anymore.
Screenshot 2024-03-15 at 10 47 54

app/workers/publish_scheduled_edition_worker.rb
```diff
@@ -0,0 +1,27 @@
+class PublishScheduledEditionWorker
+  include Sidekiq::Worker
+
+  def perform(edition_id, user_id)
+    edition = TravelAdviceEdition.find(edition_id)
+
+    if edition.scheduled_publication_time <= Time.zone.now
+      notifier.put_content(edition)
```
lauraghiorghisor-tw, line 8: We are now doing a `put_content` when we save the draft at the point of scheduling (see `schedule_for_publication` on the `TravelAdviceEdition` model) -> we don't need it here anymore, but also doesn't hurt to have it?

ryanb-gds: I don't think we should be doing this in either place, should we? In my head put content should happen only when the edition content is saved? It might be that the existing code isn't really working how I feel it should though.

In any case, I don't think "not hurting" is sufficient justification for making an API call 😉

app/workers/publish_scheduled_edition_worker.rb
ryanb-gds: If it could, then maybe it should 👀

spec/factories.rb
```diff
@@ -18,6 +18,15 @@
   factory :draft_travel_advice_edition, parent: :travel_advice_edition do
   end
 
+  factory :scheduled_travel_advice_edition, parent: :travel_advice_edition do
+    scheduled_publication_time { 1.hour.from_now }
```
ryanb-gds, line 22: Shouldn't we do this in the `after` callback as well? It doesn't make sense for the time to be set but the state not to be scheduled, right?

spec/models/travel_advice_edition_spec.rb
```diff
@@ -365,6 +400,88 @@
...
   end
 
+  context "scheduling" do
+    let!(:published) do
+      create(:published_travel_advice_edition, country_slug: "aruba", published_at: 3.days.ago, change_description: "Stuff changed")
+    end
+    let!(:scheduled) { create(:scheduled_travel_advice_edition, country_slug: "aruba") }
+
+    before do
+      published.reload
```
ryanb-gds, line 410: This... doesn't feel that it should be necessary. Wonder what bit of shared state is causing that problem 🤔

ryanb-gds (changes requested):
There's a lot here, I think it would be great to get on a call this afternoon and to talk through it all. Looking good though

app/controllers/application_controller.rb
```diff
@@ -24,4 +24,8 @@ def set_authenticated_user_header
   def skip_slimmer
     response.headers[Slimmer::Headers::SKIP_HEADER] = "true"
   end
+
+  def can_schedule_edition?
+    current_user.has_permission?(User::SCHEDULE_EDITION_PERMISSION)
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: It is done https://trello.com/c/1TKIO84l/2374-travel-advice-publisher-improvements-remove-feature-flag-and-release

app/models/travel_advice_edition.rb
```diff
@@ -153,10 +175,17 @@ def first_version?
     version_number == 1
   end
 
+  def has_valid_change_description_for_scheduling?
+    return true unless !is_minor_update? && !change_description.presence
+
+    errors.add(:change_description, "can't be blank on schedule")
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: We have a similar `We had some problems publishing: Change description can't be blank on publish.`

spec/controllers/admin/schedulings_controller_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      expect(@edition.reload.state).to eq("scheduled")
+      expect(@edition.reload.scheduled_publication_time).to eq Time.zone.local(*scheduling_params.values)
+    end
+
+    it "redirects to countries index page with success message" do
+      post :create, params: { edition_id: @edition.id, scheduling: generate_scheduling_params(3.hours.from_now) }
+
+      expect(response).to redirect_to admin_countries_path
+      expect(flash[:notice]).to eq "#{@country.name} travel advice is scheduled to publish on #{3.hours.from_now.strftime('%B %d, %Y %H:%M %Z')}."
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: We had those assertions on the queue size here, moved them in the model tests. Perhaps a single assertion on the queue size can live here and then the model tests can cover the args in more detail?

app/models/travel_advice_edition.rb
lauraghiorghisor-tw: So there are no on save callbacks or anything here, the put content calls on save draft are actioned from the edition controller when saving a simple draft. 

Our scheduled state is still a draft, and when we save nothing gets sent to publishing-api. If we want the preview links to point to an updated draft content store version, then we need to do a put content.

app/workers/publish_scheduled_edition_worker.rb
lauraghiorghisor-tw: I cannot quite think whether or not any potential changes of edition would need to be sent once you make the first put_content call on schedule. I would assume NOT, since the edition is not editable whilst scheduled?

app/assets/stylesheets/date_time.scss
```diff
@@ -0,0 +1,22 @@
...
+
+.app-c-datetime-fields__date-time-input--large {
+    max-width: 7.5em;
+    min-width: 7.5em;
+}
+
+.app-c-datetime-fields__time-separator {
+    margin: govuk-spacing(6) + 5 govuk-spacing(2) 0 0;
+    margin-bottom: 0;
+}
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: I did not try it without them. WH overrides them 😶‍🌫️

app/models/travel_advice_edition.rb
```diff
@@ -205,6 +236,10 @@ def first_version_cant_be_minor_update
     end
   end
 
+  def validate_scheduled_publication_time
+    errors.add(:scheduled_publication_time, "can't be in the past") if scheduled_publication_time && state == "draft" && scheduled_publication_time <= Time.zone.now
+  end
```
ryanb-gds, line 241: I think you could have applied this validation rule in the state machine config for the draft state (like for the can't edit in published state validation) if you wanted. Then you could avoid the `== draft` check here. Minor point though, not required to merge

ryanb-gds (approved):
Great work @lauraghiorghisor-tw and @baisa - exciting to see this ready to merge! 🚀

ryanb-gds (approved):
🤖

PR #1843, opened 2024-03-20
Cancel scheduling

lauraghiorghisor-tw:
[Trello card](https://trello.com/c/sr1PGDJf/2373-travel-advice-publisher-improvements-cancel-scheduling-flow)

### DIFF
--- a/app/workers/scheduled_publishing_worker.rb
+++ b/app/workers/scheduled_publishing_worker.rb
@@ -15,6 +15,11 @@ class ScheduledPublishingWorker
       return
     end
 
+    unless edition.scheduled? && edition.scheduled_publication_time
+      Sidekiq.logger.warn("Publishing cancelled for edition of ID '#{edition.id}'.")
+      return
+    end
+
     if edition.scheduled_publication_time > Time.zone.now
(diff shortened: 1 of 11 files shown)


app/workers/scheduled_publishing_worker.rb
```diff
@@ -15,6 +15,11 @@ def perform(edition_id)
       return
     end
 
+    unless edition.scheduled? && edition.scheduled_publication_time
+      Sidekiq.logger.warn("Edition of ID '#{edition.id}' has not been published.")
```
ryanb-gds, line 19: I'd probably like this log message to mention that the publishing has been cancelled, what do you think?

ryanb-gds (commented):
Looking good to me, though would appreciate @baisa reviewing too

One question about log message

app/workers/scheduled_publishing_worker.rb
lauraghiorghisor-tw: @baisa I think I mentioned this to you, whether there is ever a case that the time isn't set in the first place. Revisiting the flow just now and that seems unlikely. So happy to accept that the only case when we run the worker and the `scheduled_publication_time` is unset is because the edition has been intentionally unscheduled.

baisa (approved):

PR #1847, opened 2024-03-21
Accessibility improvements for scheduling flow

lauraghiorghisor-tw:
[Trello card](https://trello.com/c/HbvGkzzl/2432-accessibility-changes-tap)

### DIFF
--- a/app/assets/stylesheets/edition_form.scss
+++ b/app/assets/stylesheets/edition_form.scss
@@ -1,9 +1,3 @@
-.edition-form__action-button-group {
-  :nth-last-child(2) {
-    margin-left: auto;
-  }
-}
-
 .edition-form__action-preview-link {
   display: none;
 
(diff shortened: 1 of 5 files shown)


baisa (approved):
it looks good, thank you @lauraghiorghisor-tw and @dnkrj

app/views/components/_datetime_fields.html.erb
```diff
@@ -44,7 +44,7 @@
     <% unless date_only && !date_heading %>
       <%= render "govuk_publishing_components/components/heading", {
         text: date_heading || "Date (required)",
-        heading_level: heading_level || 3,
```
dnkrj, line 47: Should we be passing in `heading_level` as an argument rather than changing it in the component?
Or is this not actually used anywhere else?

dnkrj (commented):
Looks good, just one comment about the datetime component

app/views/components/_datetime_fields.html.erb
lauraghiorghisor-tw: Not used anywhere else, but fair shout ✅

PR #1848, opened 2024-03-21
Implement feature tests for scheduling flows

lauraghiorghisor-tw:
[trello card](https://trello.com/c/QPf16ia4/2433-feature-tests-for-tap)

### DIFF
--- a/app/models/action.rb
+++ b/app/models/action.rb
@@ -36,7 +36,7 @@ class Action
 
   def to_s
     if request_type == SCHEDULE_FOR_PUBLICATION
-      string = "Scheduled for publication"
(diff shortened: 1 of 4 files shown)


app/models/action.rb
```diff
@@ -36,7 +36,7 @@ def status_action?
 
   def to_s
     if request_type == SCHEDULE_FOR_PUBLICATION
-      string = "Scheduled for publication"
+      string = "Schedule for publication"
```
minhngocd, line 39: might be worth adding a comment in the commit message to explain this change?

minhngocd (approved):
Looks good! looks like all happy path covered 👍

PR #1869, opened 2024-04-09
Raise error when there are overdue scheduled editions

lauraghiorghisor-tw:
We want to publish any scheduled editions left unpublished. If the scheduling has somehow failed, we address that by running a  cronjob every morning, which attempts to publish the editions by executing the worker synchronously.

If any editions are still left unpublished after running the worker, we raise a an error to ensure the team has visibility. The intention is to observe this and gauge whether we should enable pager duty and/or increase the frequency of the cronjob.

 [Trello card](https://trello.com/c/ESWx6uYN/2447-missed-scheduling-jobs-log-and-alert)

### DIFF
--- a/lib/tasks/publish_scheduled_editions.rake
+++ b/lib/tasks/publish_scheduled_editions.rake
@@ -1,6 +1,17 @@
 desc "Cronjob running daily to catch potentially unpublished editions"
 task publish_scheduled_editions: :environment do
-  TravelAdviceEdition.with_state(:scheduled).pluck(:id).each do |id|
+  TravelAdviceEdition.scheduled.pluck(:id).each do |id|
     ScheduledPublishingWorker.new.perform(id.to_s)
+  rescue StandardError
+    next
+  end
+
+  overdue_editions = TravelAdviceEdition.due_for_publication
+  raise ScheduledEditionsOverdueError, overdue_editions if overdue_editions.any?
+end
+
+class ScheduledEditionsOverdueError < StandardError
+  def initialize(editions)
+    super("The following editions are due for publication: #{editions.map(&:_id).to_sentence}")
   end
 end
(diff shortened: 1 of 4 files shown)


ryanb-gds (commented):
Just looking for a quick sense check on the unhappy path for the `ScheduledPublishingWorker` - might be absolutely fine, I don't know enough about running Sidekiq jobs synchronously

lib/tasks/publish_scheduled_editions.rake
```diff
@@ -1,6 +1,11 @@
 desc "Cronjob running daily to catch potentially unpublished editions"
 task publish_scheduled_editions: :environment do
-  TravelAdviceEdition.with_state(:scheduled).pluck(:id).each do |id|
+  TravelAdviceEdition.scheduled.pluck(:id).each do |id|
     ScheduledPublishingWorker.new.perform(id.to_s)
```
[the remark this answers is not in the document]
lauraghiorghisor-tw: Implemented. I think yesterday we were thinking the publishing might go wrong, well, that's actually in its own worker so it's unlikely. But the state machine can be invalid and there is a task order validation that raises. SO it can indeed fail.

spec/models/travel_advice_edition_spec.rb
```diff
@@ -267,6 +267,15 @@
     expect(TravelAdviceEdition.published.to_a.sort).to eq([e2, e4].sort)
   end
 
+  it "has a scheduled scope" do
```
ryanb-gds, line 270: Bit of a nitpick, but I feel it would be nice if the test name explained what the scheduled scope does, especially as we're testing the order. Feels a bit implicit

spec/workers/scheduled_publishing_worker_spec.rb
```diff
@@ -1,6 +1,6 @@
 RSpec.describe ScheduledPublishingWorker, type: :worker do
   describe "#perform" do
-    let(:country) { Country.find_by_slug("afghanistan") }
+    let(:country) { Country.find_by_slug("aruba") }
```
ryanb-gds, line 3: Did you need to change this for a reason?

ryanb-gds (approved):
I think we're there. Probably would recommend making the test name a bit more explicit but looking good 👍

PR #1876, opened 2024-04-17
Rename Micronesia

GDSNewt:
Replace 'Micronesia' with 'Federated States of Micronesia' - content request [Trello](https://trello.com/c/SU6BliYA/2438-replace-micronesia-with-federated-states-of-micronesia-content-request)

The change to the 'cannot_edit_published' and 'cannot_edit_archived' methods are there to allow us to change the country slug. Prior to this change validation would have stopped us from being able to do that, and consequently all the Travel Editions would remain associated with the old slug and not be visable within the travel advice publisher UI

### DIFF
--- a/app/models/travel_advice_edition.rb
+++ b/app/models/travel_advice_edition.rb
@@ -222,13 +222,13 @@ private
   end
 
   def cannot_edit_published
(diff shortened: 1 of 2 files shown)


lauraghiorghisor-tw (approved):
Happy to approve, as we've already discussed. I would suggest deploying in integration, running the reslug, ensure all your editions are behaving as expected and if they don't, potentially write up missing redirect functionality.

PR #1879, opened 2024-04-20
Fix invalid date behaviour

lauraghiorghisor-tw:
During testing it was discovered that the generic date incorrect error was showing for seemingly correct dates.

The previous implementation made use of an Integer conversion to ascertain whether the input is a correct date/time value. Nonetheless, Integer() will interpret input such as "0_" as octal rather than decimal, so values like "08" or "09" will not be allowed. This was introduced during refactoring and not caught by tests.

We replaced the Integer conversion with `.to_i` and added some additional tests (valid date input rather than covering invalid scenarios). We also brought back the digit number regex checks, to restrict day, month, hour and minute input to 2 digits, so that a month like "0010" is invalid.

[Trello card](https://trello.com/c/ttLGDysr/2519-invalid-date-behaviour-in-tap)

error

### DIFF
--- a/app/controllers/admin/schedulings_controller.rb
+++ b/app/controllers/admin/schedulings_controller.rb
@@ -37,10 +37,11 @@ private
     raise StandardError, "cannot be blank" if scheduling_params.values.any?(&:blank?)
 
     begin
-      year, month, day, hour, minute = scheduling_params.to_h.sort.map { |_, v| Integer(v) }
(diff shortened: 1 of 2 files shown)


baisa (approved):
Good job fixing this. Maybe we can refer to (this) [https://medium.com/technological-singularity/ruby-and-octal-literals-8e08e413cae2#:~:text=In%20Ruby%2C%20leading%20zeros%20still,are%20not%20valid%20octal%20digits.] article in the PR description to give people more context?

PR #2221, opened 2025-05-02
Fix travel advice email alert

minhngocd:
Fix the case where there is a send_alert? check that relies on the edition being published within the publishing app.
At the moment in scheduled publishing worker, it queues a list of tasks before the edition model is published. This is in contrast to the edition controller where it is published as a model before being sent to publish on publishing API and email alerts. This means the send_alert check when sending email finds the edition to not be in published state on scheduling and hence does not send out any emails.

This PR
- fixes order of publish model, publish publishing API, email alert sending within the scheduled publishing worker
- add tests
- remove published state check when sending email alert as there is already an additional check that forces send_email to always follow a publishing API publish


spec/workers/scheduled_publishing_worker_spec.rb
```diff
@@ -19,6 +22,27 @@
...
+
+      expect(PublishingApiWorker.jobs.size).to eq(1)
+      travel_to(1.hour.from_now)
+      PublishingApiWorker.drain
+      expect(PublishingApiWorker.jobs.size).to eq(0)
+
+      expect(EmailAlertApiWorker.jobs.size).to eq(1)
+      travel_to(1.hour.from_now)
+      EmailAlertApiWorker.drain
+      expect(EmailAlertApiWorker.jobs.size).to eq(0)
```
ChrisBAshton, line 41: From a run through of the code, I'm not sure the multiple `travel_to`s are necessary for this to pass? I also don't see much value in verifying that the job sizes drop to 0, since `drain` is naturally going to mean that's always true, and indeed we're already not checking that for `ScheduledPublishingWorker`.

I had a think about whether the sequencing / dependencies could be made any clearer - what do you think?

```suggestion
      # Queue up and execute the ScheduledPublishingWorker
      ScheduledPublishingWorker.enqueue(edition)
      travel_to(1.hour.from_now)
      ScheduledPublishingWorker.drain

      # Check that the above created a PublishingApiWorker, then execute it
      expect(PublishingApiWorker.jobs.size).to eq(1)
      PublishingApiWorker.drain

      # Check that the above created a EmailAlertApiWorker, then execute it
      expect(EmailAlertApiWorker.jobs.size).to eq(1)
      EmailAlertApiWorker.drain
```

app/notifiers/publishing_api_notifier.rb
```diff
@@ -38,7 +38,7 @@ def publish_index
   end
 
   def send_alert(edition)
-    return unless send_alert?(edition)
+    return if edition.is_minor_update?
```
ChrisBAshton, line 41: Great 👍 but a couple of typos in the commit message (eg "Publishnig API") and it would also be nice to link (in the commit) to the existing check that renders redundant the one we're removing here.

ChrisBAshton (approved):
Nice one! Some suggestions below, nothing blocking merge.

spec/workers/scheduled_publishing_worker_spec.rb
minhngocd: yup i'll rebase it in

PR #2299, opened 2025-07-23
Remove Maslow dependency and 'user needs' association (WHIT-2019)

ChrisBAshton:
This hasn't been in active use for several years.
We’ve already announced we’re retiring Maslow, [in July 2023](https://3.basecamp.com/4322319/buckets/15005645/messages/6324550361), and had also committed to it as a [key result in Q1 2024](https://docs.google.com/presentation/d/1iKMNB53THavB_hKXYJuyJcYtONm6_va5m_QR13zI2TM/edit#slide=id.g2c57cdc01a5_3_579). We’ve already [removed](https://github.com/alphagov/info-frontend/blob/main/docs/adr/retiring-info-frontend.md) the /info/ pages, so Maslow no longer actually publishes anything that users can see.

Maslow’s continued existence is [recognised tech debt](https://trello.com/c/ARS6215q/21-maslow-publisher-exists-but-isnt-an-active-part-of-government-publishing). We therefore need to start removing the Maslow dependency from all apps that continue to refer to it.

---

### DIFF
--- a/app/presenters/links_presenter.rb
+++ b/app/presenters/links_presenter.rb
@@ -7,7 +7,6 @@ class LinksPresenter
     {
       links: {
         parent: [TravelAdvicePublisher::INDEX_CONTENT_ID],
-        meets_user_needs: [TravelAdvicePublisher::NEED_CONTENT_ID],
         primary_publishing_organisation: [TravelAdvicePublisher::PRIMARY_ORG_CONTENT_ID],
         organisations: [TravelAdvicePublisher::PRIMARY_ORG_CONTENT_ID],
       },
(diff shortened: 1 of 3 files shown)


ryanb-gds (approved):
Looks good, assuming that `meets_user_needs` links are optional in the publishing schema

ChrisBAshton:
@ryanb-gds - should be, if ChatGPT isn't leading me astray!

1. It is [nested inside the links object](https://github.com/alphagov/publishing-api/blob/553be1b2a1662ef60bf3d52add57e241171f429f/content_schemas/dist/formats/answer/frontend/schema.json#L90).
2. The links object does not have a required key — meaning none of its properties are required, including meets_user_needs.
3. Therefore, even if meets_user_needs is defined in properties, it does not need to be present in a valid document.

